PR #25, opened 2025-11-22
Allow to configure time to show in 24 hour format

git-ari:
## What changed?

Allow to configure the time in the day view to show in 12 hour or 24 hours. Added tests for this.

## Type of Change

- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [X] Changes are tested

### DIFF
--- a/src/components/event-form/event-form.test.tsx
+++ b/src/components/event-form/event-form.test.tsx
@@ -23,24 +23,6 @@ const renderEventForm = (
   )
 }
 
-// Helper function for rerender with new provider props
-const rerenderEventForm = (
-  rerender: ReturnType<typeof render>['rerender'],
-  props: Parameters<typeof EventForm>[0],
-  providerProps = {}
-) => {
-  rerender(
-    <CalendarProvider
-      events={[]}
-      dayMaxEvents={5}
-      firstDayOfWeek={0}
-      {...providerProps}
-    >
-      <EventForm {...props} />
-    </CalendarProvider>
-  )
-}
-
 describe('EventForm', () => {
   const mockOnAdd = mock(() => {})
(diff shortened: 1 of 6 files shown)


src/features/calendar/components/day-view/day-time-col.tsx
```diff
@@ -8,9 +8,13 @@ const hours = Array.from({ length: 24 }, (_, i) => i).map((hour) => {
 
 interface DayTimeColProps {
   className?: string
+  is24Hour?: boolean
 }
 
-export const DayTimeCol: React.FC<DayTimeColProps> = ({ className }) => {
+export const DayTimeCol: React.FC<DayTimeColProps> = ({
+  className,
+  is24Hour = false,
```
kcsujeet: **suggestion:** 

This can directly be accessed via `useCalendarContext` instead of having to be passed via props.

src/components/event-form/event-form.test.tsx
```diff
@@ -23,6 +23,24 @@ const renderEventForm = (
   )
 }
 
+// Helper function for rerender with new provider props
+const rerenderEventForm = (
```
kcsujeet: **suggestion:** 

I don't think this is needed. You can create a separate test case where you call `renderEventForm` with different props again.

kcsujeet:
Small update: 

Renaming the prop to `timeFormat: '12-hour' | '24-hour` so that one day we can support more formats like "12h", "24h", "auto", or locale-driven. Seems like most other libraries do this. : )

PR #56, opened 2026-01-05
feat: add HiddenHeaderButtons configuration to control visibility of header buttons in the calendar

Chiliseo:
## What changed?

Refactored hiddenHeaderButtons to use a shared HiddenHeaderButtons type instead of inline definitions. The type is defined in src/features/calendar/types/index.ts and used across:
IlamyCalendarProps and IlamyResourceCalendarProps
CalendarProviderProps and CalendarContextType
ResourceCalendarProviderProps and ResourceCalendarContextType
UseIlamyResourceCalendarContextReturn
DemoCalendarSettingsProps

## Type of Change

- [ ] 🐛 Bug fix
- [ X] ✨ New feature
- [ ] 💥 Breaking change
- [ X] 📚 Documentation
- [X ] 🔧 Maintenance

## Checklist

- [X] CI passes locally (`bun run ci`)
- [X] Changes are tested
- [X] Code follows project standards

### DIFF
--- a/src/components/demo/demo-calendar-settings.tsx
+++ b/src/components/demo/demo-calendar-settings.tsx
@@ -15,6 +15,7 @@ import {
 	SelectTrigger,
 	SelectValue,
 } from '@/components/ui/select'
+import type { HiddenHeaderButtons } from '@/features/calendar/types'
 import dayjs from '@/lib/configs/dayjs-config'
 import type { CalendarView, TimeFormat } from '@/types'
 import { ModeToggle } from './mode-toggle'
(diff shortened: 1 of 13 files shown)


kcsujeet:
Hi. Thanks for raising the PR. I think it'd be easier to hide the needed buttons with `CSS`. Is that not possible for your use case?

Chiliseo:
Yes, it's possible, but the functionality still exists. Perhaps if you can unlock the feature for clients, you won't need to manage the accessories.

PR #97, opened 2026-03-20
fix: move prev next date callback out of updater

sumanthneerumalla:
## What changed?

This changes prev/next navigation so ilamy no longer calls `onDateChange` from inside the `setCurrentDate` updater.

- compute `newDate` outside the updater
- call `setCurrentDate(newDate)`
- call `onDateChange?.(newDate)` after that

This avoids the React warning path while keeping prev/next navigation behavior and callback semantics the same from the caller's point of view.

It also adds a stock-header regression test that renders the calendar with parent-owned date state, clicks the built-in prev/next navigation, and verifies the React warning is not emitted.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

### DIFF
--- a/src/components/header/base-header.test.tsx
+++ b/src/components/header/base-header.test.tsx
@@ -1,5 +1,6 @@
 import { describe, expect, it, mock } from 'bun:test'
-import { fireEvent, render, screen } from '@testing-library/react'
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { useState } from 'react'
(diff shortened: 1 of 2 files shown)


kcsujeet:
This has been addressed here: https://github.com/kcsujeet/ilamy-calendar/pull/110

PR #113, opened 2026-04-08
Feature: Days-only option for weekly view in resource calendar

sebvg:
## What changed?

Add showHoursOnWeekView prop to resource calendar
Adds a showHoursOnWeekView boolean prop to IlamyResourceCalendar that controls whether the week view uses hourly granularity or a simpler day-based layout. When disabled, both horizontal and vertical orientations render one column/row per day instead of per hour, making it easier to display resource availability at a daily level without the clutter of hourly slots.

Screenshots of the new views below:

image

image

## Type of Change

- [ ] 🐛 Bug fix
- [x ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested


src/components/vertical-grid/vertical-grid-col.tsx
```diff
@@ -65,7 +68,7 @@ const NoMemoVerticalGridCol: React.FC<VerticalGridColProps> = ({
 								: `vertical-cell-${dateStr}-${hourStr}-00${resourceId ? `-${resourceId}` : ''}`
 						return (
 							<div
-								className="h-[60px] border-b border-r"
+								className={`h-[${cellHeight}px] border-b border-r`}
```
kcsujeet: **suggestion:** 

use `cn` for conditional classes

src/features/resource-calendar/components/week-view/resource-week-vertical.tsx
```diff
@@ -54,46 +55,105 @@ export const ResourceWeekVertical: React.FC = () => {
...
-			renderCell: (date: Dayjs) => (
-				<div className="text-muted-foreground p-2 text-right text-[10px] sm:text-xs flex flex-col items-center">
-					{date.format(timeFormat === '12-hour' ? 'h A' : 'H')}
-				</div>
-			),
-		}),
-		[hours, timeFormat]
+		() =>
+			showHoursOnWeekView
+				? {
```
kcsujeet: **suggestion:** 

Some properties are almost same. This can be simplified like:

```
id: showHoursOnWeekView ? 'time-col' : 'date-col'
```

src/components/vertical-grid/vertical-grid-col.tsx
kcsujeet: **issue:** 

dynamic classes will not work. Tailwind doesn't know the `cellheight` value.

src/features/resource-calendar/components/week-view/resource-week-vertical.tsx
```diff
@@ -159,7 +219,7 @@ export const ResourceWeekVertical: React.FC = () => {
 						</span>
 					</div>
 					{columns.map((col, index) => {
-						const day = col.day
+						const day = col.day as Dayjs
```
kcsujeet: **issue:** Type assertion bypasses safety here. After the early return, `col.day` can still be `undefined` since the `columns` array mixes column shapes. Use a runtime guard instead:

```tsx
const day = col.day
if (!day) return null
```

src/features/resource-calendar/types/index.ts
```diff
@@ -33,6 +33,7 @@ export interface IlamyResourceCalendarProps
 	 * - "vertical": Resources are columns, time is rows
 	 */
 	orientation?: 'horizontal' | 'vertical'
+	showHoursOnWeekView?: boolean
```
kcsujeet: **suggestion:** Missing JSDoc — the existing `orientation` prop is well-documented. For public API consistency:

```tsx
/**
 * Whether to show hourly time slots in the week view.
 * When false, renders one column/row per day instead of per hour.
 * @default true
 */
```

Also consider `weekViewGranularity: 'hourly' | 'daily'` — avoids double-negative confusion and is easier to extend later.

src/features/resource-calendar/components/week-view/resource-week-vertical.tsx
```diff
@@ -54,46 +55,105 @@ export const ResourceWeekVertical: React.FC = () => {
...
+			resources,
+			weekDays,
+			businessHours,
+			hideNonBusinessHours,
+			visibleDays.map,
+			showHoursOnWeekView,
+		]
 	)
 
+	if (!showHoursOnWeekView) {
```
kcsujeet: **suggestion:** The two return paths duplicate the `VerticalGrid` shell and header structure (sticky spacer + resource cells). Consider extracting the shared wrapper and only branching on the parts that differ (columns definition + header content) to avoid maintaining two copies.

src/components/vertical-grid/vertical-grid-col.tsx
```diff
@@ -24,6 +24,8 @@ export interface VerticalGridColProps {
 	cellSlots?: number[]
 	/** Whether this is the last column in the grid */
 	isLastColumn?: boolean
+	/** Height of each cell in pixels (default: 60) */
+	cellHeight?: number
```
kcsujeet: **issue:** `cellHeight` is added to the shared `VerticalGridColProps` interface but only the resource week view uses it. The `cellSlots` branch (line 88) still hardcodes `h-[15px]` and uses the broken dynamic class, ignoring `cellHeight`. Either make it work consistently across both branches or keep it scoped to the resource view.

src/features/resource-calendar/components/week-view/resource-week-vertical.tsx
```diff
@@ -54,46 +55,105 @@ export const ResourceWeekVertical: React.FC = () => {
...
+						days: weekDays,
+						gridType: 'day' as const,
+						cellHeight: 72,
+					})),
+		[
+			resources,
+			weekDays,
+			businessHours,
+			hideNonBusinessHours,
+			visibleDays.map,
```
kcsujeet: **issue (pre-existing):** `visibleDays.map` references `Array.prototype.map` — a stable function reference that never changes. This should be `visibleDays` to properly re-memoize when the array contents change.

src/features/resource-calendar/components/week-view/resource-week-vertical.tsx
```diff
@@ -54,46 +55,105 @@ export const ResourceWeekVertical: React.FC = () => {
...
+				? {
+						id: 'time-col',
+						days: hours,
+						day: undefined,
+						className:
+							'shrink-0 w-16 min-w-16 max-w-16 sticky left-0 bg-background z-20',
+						gridType: 'hour' as const,
+						noEvents: true,
+						renderCell: (date: Dayjs) => (
+							<div className="text-muted-foreground p-2 text-right text-[10px] sm:text-xs flex flex-col items-center">
```
kcsujeet: **suggestion:** Both `renderCell` branches (here and line 83) use the identical wrapper classes: `"text-muted-foreground p-2 text-right text-[10px] sm:text-xs flex flex-col items-center"`. Extract the shared properties and only vary the differing ones inline — same applies to `day: undefined`, `className`, and `noEvents: true` which are duplicated across both branches of this ternary.

kcsujeet (changes requested):
No test files were changed. This adds a new rendering mode with two distinct code paths — it should have test coverage for at least: default behavior preserved, days-only mode renders correctly in both orientations, and header structure differences.

kcsujeet:
Hi @sebvg Thnks for the PR. This is now available in [v1.6.1](https://github.com/kcsujeet/ilamy-calendar/releases/tag/v1.6.1)

PR #122, opened 2026-05-12
feat: render now-line in horizontal resource hour grids (closes #121)

### LINKED ISSUE #121: The "now" line is missing from horizontal views
Today the red current-time line appears in vertical day/week views, but
never in horizontal resource layouts. The only cue for "where are we
right now?" is a faint blue tint on the current hour's header cell —
which is easy to miss if users arent aware that it exists.

For a user scanning a long horizontal grid, that's the difference
between an event that **"starts in 5 minutes"** vs **"started 50 minutes ago"**.

## What we saw in the code

- The `CurrentTimeIndicator` component is only mounted by the *vertical*
  events layer. Horizontal grids never render it.
- The component itself positions with `top: {progress}%` — wrong axis
  for horizontal anyway, so it couldn't just be dropped in as-is.
- The `renderCurrentTimeIndicator` prop is accepted on the calendar,
  but never gets called when `orientation="horizontal"`. Passing it
  silently does nothing.

## Repro

```tsx
<IlamyResourceCalendar
  orientation="horizontal"
  renderCurrentTimeIndicator={({ progress }) => (
    <div style={{ left: `${progress}%`, position: "absolute", inset: "0 auto 0 0", width: 2, background: "red" }} />
  )}
/>
```

The render function is never called. Nothing draws.

## A direction that should work

Mirror the existing component for horizontal layouts:

1. Compute `progress` across the row's visible time range.
2. Render absolutely-positioned inside the horizontal row at
   `left: {progress}%`, full row height.
3. If the consumer passed `renderCurrentTimeIndicator`, call it —
   otherwise draw a default vertical line.
4. Hide when the current time falls outside the row's range.

`RenderCurrentTimeIndicatorProps` already exposes `progress: number`
(not `top` or `left`), so the existing type works for both axes —
consumers just position with `left` instead of `top`. No type changes
needed.

## Suggested file changes

Bias toward parameterizing the existing `CurrentTimeIndicator` over
forking a horizontal sibling — the only real difference between axes
is which dimension carries the `progress%`. Keeping a single component
also lets the existing test file cover both modes.

1. **`src/components/current-time-indicator.tsx`** — the component
   that draws the line. Currently always positioned vertically via
   `top: {progress}%`.

   Add an `axis?: 'vertical' | 'horizontal'` prop (default
   `'vertical'`). The default-render branch switches on `axis`:
   vertical keeps today's `top: {progress}%` + `left-0 right-0 h-0.5`;
   horizontal becomes `left: {progress}%` + `top-0 bottom-0 w-0.5`.

   Custom-render delegation via `renderCurrentTimeIndicator` is
   unchanged — the props it receives (`progress`, `currentTime`,
   `rangeStart`, `rangeEnd`, `resource`, `view`) are already
   axis-agnostic.

2. **`src/components/horizontal-grid/horizontal-grid-events-layer.tsx`**
   — the events layer that owns the row's absolutely-positioned
   children in horizontal resource views. Currently never mounts the
   indicator.

   Add `resource?: Resource` to the props interface, compute
   `rangeStart = days.at(0)` and
   `rangeEnd = days.at(-1)?.add(1, 'hour')`, then render
   `<CurrentTimeIndicator axis="horizontal" rangeStart rangeEnd resource />`
   gated on both being truthy. Mirrors the existing block in the
   vertical events layer one-for-one.

3. **`src/components/horizontal-grid/horizontal-grid-row.tsx`** — the
   row component that owns the per-row events layer. It already has
   `resource` via its own props; today it doesn't pass it down.

   Thread `resource` through to `HorizontalGridEventsLayer` so the
   custom-render callback can scope its rendering per-resource (same
   way the vertical path already does).

4. **`src/components/current-time-indicator.test.tsx`** — the
   existing test file for the indicator.

   Extend with `axis="horizontal"` cases: default markup uses `left:`
   and `w-0.5`, the indicator hides when `now` is outside range, and
   the custom-render delegate still receives a valid `progress`
   number.

No public-API changes — `RenderCurrentTimeIndicatorProps`,
`IlamyCalendarProps.renderCurrentTimeIndicator`, and the existing
context wiring stay as-is.

---

Happy to take a stab at the PR — just want to make sure this issue is communicated and the community is ok with the direction

sumanthneerumalla:
## What changed?

Closes #121.

Adds the "now" line to horizontal resource hour grids. Previously the
indicator only rendered in vertical day/week grids via
`VerticalGridEventsLayer`; this PR mirrors that mount in
`HorizontalGridEventsLayer`.

**DRY**, per your suggestion — parameterized the existing
`CurrentTimeIndicator` with an optional `axis?: 'vertical' | 'horizontal'`
prop (default `'vertical'`). The default-render branch switches
`top`/`left` and `h-0.5`/`w-0.5` based on axis. Custom-render callbacks
receive `axis` so consumers can branch on it.

The indicator now mounts in Resource Day Horizontal and Resource Week
Horizontal (hourly granularity). Gated to skip regular MonthView, per
your comment.

### Public API

Added an optional `axis?: 'vertical' | 'horizontal'` field to
`RenderCurrentTimeIndicatorProps`. The library always populates it, so
consumers can treat it as always defined inside their render function.
Strict-mode TS consumers that construct the props as literals stay
unaffected. Updated the `@example` in
`IlamyCalendarProps.renderCurrentTimeIndicator` to show the horizontal
branch.

## Type of Change

- [x] ✨ New feature
- [x] 📚 Documentation (updated `@example` and docstrings)

## Checklist

- [x] Changes are tested (2 `axis="horizontal"` cases in
      `current-time-indicator.test.tsx`, plus an `axis: 'vertical'`
      default assertion on the existing `passes correct props` test;
      2 gate integration tests in `horizontal-grid-row.test.tsx`)


src/features/calendar/types/index.ts
```diff
@@ -67,6 +67,18 @@ export interface RenderCurrentTimeIndicatorProps {
...
+	 * - `'vertical'`: `progress` maps to a `top` offset (day/week vertical grids).
+	 * - `'horizontal'`: `progress` maps to a `left` offset (horizontal resource hour grids).
+	 *
+	 * Lets a single render function position correctly across both grid layouts.
+	 *
+	 * Note: distinct from the `orientation` prop on `IlamyResourceCalendar`.
+	 * `orientation` chooses which view component renders the calendar; `axis` tells
+	 * your render function which dimension `progress` maps to inside that view.
+	 */
+	axis: 'vertical' | 'horizontal'
```
kcsujeet: suggestion (blocking): Make `axis` optional rather than required.

Issue #121 explicitly states "No public-API changes." Optional gives the same UX (the library always passes the value), aligns with the issue's intent, and avoids a hard compile break for TS strict-mode consumers who construct the props as a literal. The component already defaults to `'vertical'` at `current-time-indicator.tsx:36`.

src/components/horizontal-grid/horizontal-grid-events-layer.tsx
```diff
@@ -1,28 +1,55 @@
...
 		<div
 			className="absolute inset-0 pointer-events-none z-10 overflow-clip"
 			data-testid={dataTestId}
 		>
+			{showNowLine && rangeStart && rangeEnd && (
+				<CurrentTimeIndicator
+					axis="horizontal"
+					rangeEnd={rangeEnd}
+					rangeStart={rangeStart}
+					resource={resource}
```
kcsujeet, line 50: issue (blocking): No integration test locks in the `gridType === 'hour'` gate.

Component-level tests verify axis behavior in isolation, but nothing asserts the indicator mounts here when `gridType === 'hour'` and not when `gridType === 'day'`. One render with `data-testid='current-time-indicator'` assertions per gridType would catch a future regression.

kcsujeet (changes requested):
Inline comments below.

sumanthneerumalla:
Both addressed in 5fc1857.

- `axis` is now optional on `RenderCurrentTimeIndicatorProps`. Library
  always populates it so consumers see a defined value; making it
  optional just stops strict-mode literal constructors from breaking,
  per the "no public-API changes" line from #121.

- Added two gate tests in `horizontal-grid-row.test.tsx` — one renders
  with `gridType='hour'` and asserts the indicator is present, one
  with `gridType='day'` and asserts it isn't.

PR body updated to drop the breaking-change framing.

kcsujeet (approved):
Awesome. Thank you for making the changes. Merging now. Will go out sometime later this week.

PR #124, opened 2026-05-15
Fix date translations and localized format

maxdelorme:
## What changed?

- it ensures that locale is taken into account when it changes (on the demo changing the locale do nothing for dayJS) (fixes #141)
- it removes any translations key related to date, it is now manage by dayJS. 
   -  commit `7fc2bf8` fixes #143
   - commit `93684da` fixed #144
- it formats properly according the locale the short , long date (especially in the week selector) fix : 
   - commit `7c49d07` fixes #142 , 
   - commit `663bc86` fixes #145, 
   - commit `663bc86` fixes #146
- it takes care of firstDayOfWeek in calendar (datePicker) and in YearView  
`ef88566` fixes #147
`10f9b6e` fixes #148
- it translates properly FirstLetter of Day of week in YearView and calendar (#147, #148)
- It ensures that the format MMM D is not used when it has nos meaning for some languages (like in French)

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] CI passes locally (`bun run ci`)
- [x] Changes are tested
- [x] Code follows project standards (I hope so)

PR #125, opened 2026-05-15
Fix perfect align week column

### LINKED ISSUE #160: [BUG] In every view grid separators are misaligned between header/all-day rows and time grid
**Bug descpription**
In every view layout, the column separators do not line up perfectly between the header/all-day area and the timed grid below.

Once the timed rows start, the vertical separator between days appears slightly thicker than the separator above in the all-day row and week/day header. The timed grid columns also appear shifted a few pixels to the left compared with the header/all-day rows, so the grid lines do not connect cleanly.

**Expected behavior**
The day column separators should align exactly from the week header through the all-day row into the timed grid. The border thickness should look consistent across all sections. Remove the bold border from the lines between the days.

Image

maxdelorme:
fixes #160 
## What changed?

Before 
Capture d’écran 2026-05-15 à 10 44 59
After
Capture d’écran 2026-05-15 à 11 09 53

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested


kcsujeet:
**Cannot review — branch is stacked on top of PR #124's branch.**

This PR's diff shows ~50 files because it includes every commit from #124 (and the chain below it), not just the changes specific to this PR's title. Please rebase this branch off `main` so it only contains the commits unique to this PR. Once rebased, the diff will be small and reviewable.

See #124 for the broader context on PR scoping.

maxdelorme:
successfully rebase to main .
Thanks for looking at it

kcsujeet (commented):
question: Could you share repro steps for the alignment issue this PR is fixing?

I tried to reproduce on the live demo at https://ilamy.dev (Week view, default settings, multiple viewport widths) and the columns line up for me. The "Before" screenshot in the PR description shows a real misalignment, so I want to make sure I am testing the same conditions you saw it under. Could you list:

- Viewport width (or whether you had devtools open at a specific size).
- Locale (the screenshot looks like French).
- First day of week, hidden days, business hours, or any other prop overrides.
- Browser.

Without a repro I cannot tell which of the changes in this PR are load-bearing for the fix and which are doing something else, which makes the rest of the review harder.

suggestion: Consider splitting the non-alignment changes into separate PRs.

Independent of the fix itself, a few things in this diff look broader than alignment and would be easier to review on their own:

- `AnimatedSection`'s default className changes from `'inline-block w-full'` to `'min-w-0'`, which affects every consumer of the component (around 10 call sites, not just the week view).
- `VerticalGridHeaderContainer`'s default className change has a similar reach across views.
- `grid-cell.tsx` drops `only:border-r` from the default class.
- `HOUR_ROW_HEIGHT_PX` is added to `constants.ts` but I could not find a consumer in this diff.
- Several shared components gain new optional props (`gutterClass`, etc.) that look like they are only used by one caller.
- The 2026-05-17 dev log mentions "scroll-to-now" work that I could not see in this diff. Is that intended for a follow-up PR, or did it land somewhere I missed?

Once we have a clear repro for the alignment issue, the minimal fix might be just one or two lines, and the rest of these can be evaluated on their own merits.

maxdelorme:
fix #160 
commit totally review and much more simpler

maxdelorme:
a good way to reproduce is to add : 
```css
  --spacing: 0.5rem;
```
in the `:root` of `global.css` file

kcsujeet:
Verified the `--spacing: 0.5rem` repro side by side. Bug on main, fixed on this branch. Nice catch.

One ask before merge: add a regression test for the alignment. Either a unit test asserting the day columns / time-gutter widths under a non-default `--spacing`, or a snapshot of the week-view's resolved layout. Otherwise the next refactor of `CELL_CLASS` silently regresses this.

maxdelorme:
closed by mistake

maxdelorme:
-

maxdelorme:
could you reopen the pull request I delete my branch by mistake and it closed the PR

kcsujeet:
@maxdelorme 
I'm unable to do that since the branch seems to be delete in origin too. I'd say the only option is to open another PR.

PR #135, opened 2026-05-17
Fix(style): height of body was not dynamic

maxdelorme:
## What changed?

The main body of the calendar was not flex, now the body take all the available height and the header what it need

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested


kcsujeet (changes requested):
Could you link a related issue and provide steps to reproduce before this can be merged?

The PR description says the body height was not dynamic, but I cannot tell from the diff alone what was broken, under what conditions, or how to verify the fix. Without a linked issue describing the bug and concrete repro steps (viewport, props, browser, what was seen versus expected), I cannot review or accept this change.

In general, please open an issue for any bug first, then reference it from the PR. That gives both of us a shared place to confirm the bug exists, agree on the expected behavior, and verify the fix afterwards. Happy to take another look once that is in place. Thanks.

maxdelorme:
I have a container with padding : 
Without min-h-0 the body as a min-h to content: 
so I got this 
image
with min-h-0 : 
The padding is respected 
image

kcsujeet:
@maxdelorme Thanks for the screenshots. Quick question: what's the height setup on your parent container?

 The calendar needs a parent with a defined height to compute its own layout. You can try it out in the demo. https://ilamy.dev/demo/
  
Screenshot 2026-05-21 at 9 43 44 PM

maxdelorme:
The parent is : 
```html
<section class="min-w-0 flex-1 flex flex-col items-center justify-start p-10  overflow-y-auto">
   <div ... id="ilamy-calendar">...</div>
</section>
```


and all other parents are full-h so section is full height. 
Using a compute value in css is most of the time a bad idea has you can not adapt to .

Step to reproduce on you demo 
1. Goes in week-view 
2. Resize width of you're browser around 1000px so the header title goes above View Switcher
3. See a height of 400px in the setting panel
4. on the div with corresponding to card-content add a padding : 
```html
<div data-slot="card-content" class="p-0 overflow-clip relative z-10" style="height: 400px;padding: 2rem;">
```
image


the padding bottom is not respected because the height is not dynamic (your header is no more 3.5rem)

kcsujeet:
Looks good. Quick ask: add a regression test for the week-view alignment. Custom headerComponent or responsive wrap case works.

Also #125 has the same commit, mind closing one?

maxdelorme:
I removed commit on #125.
> Quick ask: add a regression test for the week-view alignment. 

I'm not sure how to do this , are you sure it is not related to #125

kcsujeet:
You're right, test belongs on #125. This PR is good as-is, will approve.

kcsujeet (approved):

PR #136, opened 2026-05-17
feat(calendar): add scrollTime prop for hour-resolution views (closes #132)

### LINKED ISSUE #132: [FEATURE] Add scrollToTime functionality
Currently, the calendar automatically scrolls to the start of businessHours (e.g., 08:00). However, our application requires 24/7 business hours (00:00 to 24:00) so users can see night shifts, but we still want the calendar to automatically scroll to 08:00 on load to focus on the main daytime slots.

Since setting businessHours to 00:00 forces the calendar to start at midnight, we need a way to decouple the initial scroll position from the business hours configuration.

**Proposed Solution**
Introduce a dedicated prop (similar to react FullCalendar's scrollTime) to explicitly set the default scroll position, regardless of what is configured in businessHours.

**JavaScript**
```
<Calendar
  businessHours={{ start: '00:00', end: '24:00' }} // Show the whole day
  scrollTime="08:00:00"                            // But focus on 8 AM on load
/>
```

By the way: Good library :-)
Thanks

kcsujeet:
Closes #132

## Summary

Adds `scrollTime?: string` (HH:mm or HH:mm:ss) to `IlamyCalendarProps`, inherited by `IlamyResourceCalendarProps`. On mount and on view/date change, the calendar scrolls its internal time grid to the row matching the configured hour. Works across all six hour-resolution views:

- Day, Week (regular calendar)
- Resource Day vertical, Resource Week vertical
- Resource Day horizontal, Resource Week horizontal

API name matches FullCalendar's `scrollTime` (verified via WebFetch against [fullcalendar.io docs](https://fullcalendar.io/docs/scrollTime)).

## How it works

- **Anchor**: a `data-hour="HH"` attribute on the time-gutter cells (vertical) and time-header cells (horizontal). Production scroll behavior is independent of `data-testid`.
- **Scroll target**: the Radix ScrollArea viewport directly via `Element.scrollTo({ top })` / `scrollTo({ left })`. `scrollIntoView` would scroll all ancestor containers including the page body.
- **Offset math**: computed relative to the first hour row (`targetRect.left - firstRect.left`). The first row naturally sits past any sticky-left column (Resources) or sticky-top header, so the target hour lands at the visible start of the time area.
- **Out-of-range**: clamps silently to the nearest visible row (works with `hideNonBusinessHours`).
- **Auto-height calendars**: silently no-ops since there is no scroll container.

## Bonus

Second commit adds a DRY guiding-principle note to the code-review skill (Agent 2). Encourages reviewers to actively hunt for repetition across production code, tests, and demos, with the recently-added `renderScrollHook` helper as the kind of pattern to look for.

## Test plan

- [ ] Day view: set `scrollTime="08:00:00"` → time grid opens at 8 AM.
- [ ] Week view: same as day.
- [ ] Resource day horizontal: 8 AM appears at the visible start of the time area (right of the Resources column).
- [ ] Resource week horizontal: same.
- [ ] Day → Week → Day navigation: each mount re-scrolls to 8 AM.
- [ ] Out-of-range with `hideNonBusinessHours` (e.g. `scrollTime="03:00"`, businessHours 8-18): clamps to first visible row.
- [ ] Page is not scrolled by the calendar scrolling internally.
- [ ] `bun run ci`: 842 tests, 0 fail.


maxdelorme:
I've worked also on autoscroll when changing view to day or week view, it autoscroll to current time

kcsujeet:
Thanks for the heads up, @maxdelorme. I do not see an open PR for that work.

What you described (auto-scrolling to the current time) is a different feature from this PR. FullCalendar's `scrollTime` is a configurable initial scroll position with a default of `06:00:00`, not a "scroll to now" anchor (https://fullcalendar.io/docs/scrollTime). And the feature request in #132 explicitly asks for that same behavior: a configurable position decoupled from `businessHours`, with `scrollTime="08:00:00"` as the example.

If you want a separate prop for auto-scrolling to the current time, please open a new issue or PR so it can be considered on its own.

maxdelorme:
You're PR is better, and goes further than mine, I will just need to pass `now` to your prop to have the equivalent, so thanks again.

Max

PR #154, opened 2026-05-18
fix(resource-calendar): sync resources prop after mount (closes #153)

### LINKED ISSUE #153: `IlamyResourceCalendar` ignores changes to the `resources` prop after mount
`<IlamyResourceCalendar resources={…}>` only reads the `resources`
prop once at mount. Any subsequent change to the prop is silently
ignored — the calendar continues to display the resource list
captured on the first render. The only workaround today is forcing
a full remount via the React `key` prop, which discards all internal
state (scroll position, current view date, visibility toggles).

In our app this surfaces when the user picks a different facility in
a facility selector — the new resource list comes back from the
server fine, but the calendar keeps displaying the previous
facility's resources.

## What we saw in the code

- `src/features/resource-calendar/contexts/resource-calendar-context/provider.tsx:84` —
  `currentResources` is seeded from `resources` in `useState` and
  never updated. No setter, no syncing effect.
- Same file, line 85-87 — `visibleResources` is initialized from
  `resources` exactly once; the only mutations are user-triggered
  show/hide callbacks.
- Downstream getters (`getResourceById`, `getVisibleResources`,
  `showAllResources`) and the context value (line 230) all read from
  `currentResources`, so every consumer sees the stale snapshot.

## Repro

```tsx
function Example() {
  const [resources, setResources] = useState([
    { id: "court-1", title: "Court #1", businessHours: defaultBH },
  ]);
  return (
    <>
      <button
        onClick={() =>
          setResources([{ id: "court-2", title: "Court #2", businessHours: defaultBH }])
        }
      >
        Swap to Court #2
      </button>
      <IlamyResourceCalendar resources={resources} events={[]} />
    </>
  );
}
```

Expected: clicking the button replaces Court #1 with Court #2.
Actual: the calendar continues to display Court #1.

## A direction that should work

1. Sync `currentResources` to the prop. Simplest form is to drop the
   internal `useState` entirely and read the prop directly — nothing
   in the provider mutates `currentResources`, so the wrapper buys
   nothing.

2. Sync `visibleResources` carefully. Two failure modes to avoid:
   - **Naive overwrite** (`setVisibleResources(new Set(resources.map(r => r.id)))`)
     resets the set on every prop change, resurrecting resources the
     user had explicitly hidden.
   - **Pure add** (`setVisibleResources(prev => new Set([...prev, ...newIds]))`)
     never drops removed resources, so stale IDs accumulate.

   The right behavior is probably: new resources default to visible,
   existing-and-currently-visible stay visible, existing-and-hidden
   stay hidden, removed resources drop out. That needs a `useRef` of
   previously-seen IDs to distinguish "new" from "hidden by user."

## Workaround today

```tsx
<IlamyResourceCalendar key={facilityId} resources={resources} … />
```

Forces a full remount each time, at the cost of losing scroll
position, current view date, visibility toggles, and any drag state.

Happy to take a stab at the PR if there's interest.

kcsujeet:
Closes #153

## Summary

`IlamyResourceCalendar` was ignoring updates to the `resources` prop after mount because `ResourceCalendarProvider` wrapped it in `useState`. Consumers had to force a full `key` remount to swap resources, which discarded scroll position, current view date, and other internal state.

## What changed

1. **Drop the `useState` snapshot.** `ResourceCalendarProvider` now reads the `resources` prop directly, so prop updates flow through immediately.

2. **Delete the dead resource-visibility API.** While auditing, the `visibleResources` / `getVisibleResources` / `showResource` / `hideResource` / `toggleResourceVisibility` / `showAllResources` / `hideAllResources` surface turned out to have zero callers outside the provider and was not exposed in `src/index.ts`. Removed the whole system. All five resource views now read `resources` from context directly. Provider drops from 330 to 274 lines.

3. **Cleaned up `CustomResourceEventForm` test fixture** in `ilamy-resource-calendar.test.tsx`: dropped three `!` non-null assertions, three `as CalendarEvent` casts, nine useless `renderEventForm={(props) => <Fixture {...props} />}` wrappers, and the unused close button.

## Test plan

- [x] New TDD tests in `ilamy-resource-calendar.test.tsx`:
  - `reflects a replaced resources prop without remount` (the #153 repro)
  - `shows newly-added resources by default`
  - `drops removed resources from the visible set`
- [x] All existing tests pass (857 pass, 0 fail).
- [x] `bun run ci` green (lint, type-check, test, build).


sumanthneerumalla:
**In-place array mutation will silently miss updates.** Now that
`resources` is read directly, a consumer doing `resources.push(...)`
on the same array reference won't trigger a re-render — same caveat
as the old code, but more visible now that the prop is the canonical
source of truth. Might be worth a one-line JSDoc on the `resources`
prop: "Pass a new array reference (not a mutation) to update."

kcsujeet:
@sumanthneerumalla Not specific to this prop or PR. React doesn't know about `push` / `pop` at all, it only knows state changes. Same caveat applies to `events`, `hiddenDays`, and every object/array prop the component takes. Documenting immutability per prop would just be noise.

Skipping the JSDoc.

PR #161, opened 2026-05-20
Fix(recurrence): prevent lost of beginning of series when editing a recurring event with scope "all" to be lost

### LINKED ISSUE #150: [BUG] When moving the whole recurrent series, events are lost
in the demo select on item of the series "Daily Standup" move it latter.
choose all (moving the whole series)
items before the one selected disappear

maxdelorme:
fixes #150 

## What changed?

It computes correctly the new beginning of the base event to ensure it is properly set and display properly

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested


src/features/recurrence/utils/recurrence-handler.ts
```diff
@@ -281,10 +326,20 @@ export const updateRecurringEvent = ({
...
+				start: _start,
+				end: _end,
+				rrule: _rrule,
+				...nonDateUpdates
+			} = updates
 			const updatedBaseEvent = {
 				...baseEvent,
-				...updates,
+				...nonDateUpdates,
+				start: anchored.start,
```
kcsujeet, line 332: suggestion: Drop the `nonDateUpdates` destructure.

The explicit `start`, `end`, and `rrule` properties at the bottom of the object literal (lines 338-340) already override anything spread before them, so `...nonDateUpdates` can be `...updates` and the 6-line destructure block on lines 329-334 deletes:

```ts
case 'all': {
  const anchored = applyAllScopeUpdates(baseEvent, updates)
  updatedEvents[baseEventIndex] = {
    ...baseEvent,
    ...updates,
    start: anchored.start,
    end: anchored.end,
    rrule: anchored.rrule,
  }
  break
}
```

Same final shape, less code.

src/features/recurrence/utils/recurrence-handler.ts
```diff
@@ -82,6 +82,51 @@ const findBaseEventIndex = (
...
+	const newEnd = newStart.add(updatedDurationMs, 'millisecond')
+
+	const mergedRrule = updates.rrule ?? baseEvent.rrule
+	const syncedRrule = mergedRrule
+		? ({ ...mergedRrule, dtstart: newStart.toDate() } as RRuleOptions)
+		: undefined
+
+	return { start: newStart, end: newEnd, rrule: syncedRrule }
+}
+
```
kcsujeet, line 127: suggestion: Helper can be smaller and more direct.

Two branches today (early-return when `!updates.start`, main path otherwise) mix three concerns: time-of-day extraction via a 4-call `.hour().minute().second().millisecond()` chain, duration math, and rrule sync. Pulling "date from base, time-of-day from source" into a tiny helper and using flat if/else for the end-time cases reads top-down as "newStart, then newEnd, then newRrule, then return":

```ts
const withTimeOfDay = (target: Dayjs, source: Dayjs): Dayjs => {
  const timeOfDayMs = source.diff(source.startOf('day'), 'millisecond')
  return target.startOf('day').add(timeOfDayMs, 'millisecond')
}

const applyAllScopeUpdates = (
  baseEvent: CalendarEvent,
  updates: Partial<CalendarEvent>
): { start: Dayjs; end: Dayjs; rrule: RRuleOptions | undefined } => {
  let newStart = baseEvent.start
  if (updates.start) {
    newStart = withTimeOfDay(baseEvent.start, updates.start)
  }

  let newEnd = baseEvent.end
  if (updates.start && updates.end) {
    newEnd = newStart.add(updates.end.diff(updates.start), 'millisecond')
  } else if (updates.start) {
    newEnd = newStart.add(baseEvent.end.diff(baseEvent.start), 'millisecond')
  } else if (updates.end) {
    newEnd = withTimeOfDay(baseEvent.end, updates.end)
  }

  const mergedRrule = updates.rrule ?? baseEvent.rrule
  let newRrule: RRuleOptions | undefined
  if (mergedRrule) {
    newRrule = { ...mergedRrule, dtstart: newStart.toDate() }
  }

  return { start: newStart, end: newEnd, rrule: newRrule }
}
```

What changed:
- `withTimeOfDay` extracted. Says what it does in its name.
- Flat if/else for the four end-time cases, no nested ternaries.
- The `as RRuleOptions` casts drop. TS narrows `mergedRrule` to `RRuleOptions` after the `if`; the spread keeps `freq` and overrides `dtstart` with `Date`. If TS still complains, the cast can come back.
- Single return at the end.

Existing tests pass either way (both PR tests send `start+end` together; the existing "should update the base recurring event" sends `{title, rrule}` with no dates and falls through to the defaults). The already-posted comment about dropping `nonDateUpdates` becomes irrelevant if this lands, since the helper returns canonical `start/end/rrule`.

src/features/recurrence/utils/recurrence-handler.ts
maxdelorme: done good catch

src/features/recurrence/utils/recurrence-handler.ts
maxdelorme: Thanks, that's clearer, I add some comments to help reader to understand what's going on

maxdelorme:
Thanks for the review.
All changes applied

kcsujeet (commented):
Thank you for the PR.

PR #169, opened 2026-05-22
Fix perfect align week column

### LINKED ISSUE #160: [BUG] In every view grid separators are misaligned between header/all-day rows and time grid

maxdelorme:
## What changed?
The following of #125 
fixes #160

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [ ] Changes are tested


maxdelorme:
I added test but not really sure if this is what you want.
I have no knowledge about regression test on IHM. But I will eager to learn

src/features/calendar/components/week-view/week-view.test.tsx
```diff
@@ -843,6 +846,29 @@ describe('WeekView', () => {
...
+
+		const firstDayColumn = screen.getByTestId(
+			`vertical-col-day-col-${firstDay}`
+		)
+		expect(firstDayColumn.className).toContain(sharedFluidWidthClass)
+
+		const firstAllDayCell = screen.getByTestId(`day-cell-${firstDay}`)
+		expect(firstAllDayCell.className).toContain(sharedFluidWidthClass)
+	})
+
```
kcsujeet: issue: The `wrapperStyle={{ '--spacing': '12px' }}` setup doesn't affect any assertion.

JSDOM doesn't compute layout from CSS variables, so the same `className.toContain('flex-1')` checks would pass with or without the wrapper style. The test name claims it verifies alignment under custom `--spacing`, but the body only asserts that specific class strings are present. Real alignment regressions (different padding, missing `min-w-0` in an inner flex child, switching the parent to grid) wouldn't fail it, while a benign refactor (renaming `flex-1` to a custom utility) would. Either drop the `wrapperStyle` plumbing and rename to reflect what's actually checked, or move to a real-browser layout test with `getBoundingClientRect`.

src/components/vertical-grid/vertical-grid-col.tsx
```diff
@@ -50,7 +50,8 @@ const NoMemoVerticalGridCol: React.FC<VerticalGridColProps> = ({
 	return (
 		<div
 			className={cn(
-				'flex flex-col flex-1 items-center justify-center min-w-20 bg-background relative',
+				'flex flex-col flex-1 items-center justify-center bg-background relative',
```
kcsujeet: question: Was dropping `min-w-20` intentional?

The previous version held a 5rem floor on column width. Without it, on viewports narrower than `(daysShown * 5rem) + gutter`, columns can shrink below that minimum and event blocks get squashed. If `min-w-0` everywhere is the new design (fully fluid week view), worth a dev-log note so this isn't re-added later. If unintentional, restore `min-w-20` alongside the `min-w-0` on flex children.

maxdelorme: I think it's better to a component to be full fluid, that's why it's removed it. 
This is to the user to ensure the size he can give it

maxdelorme:
tests rewrite, I think is much better

src/components/vertical-grid/vertical-grid-col.tsx
kcsujeet: @maxdelorme Not having a `min-width-20` breaks the layout. Please test changes in both regular and resource calendar in all modes in various screen sizes. 

Screenshot 2026-05-24 at 2 43 17 PM

maxdelorme: I restore the min-w-20
and resolve another bug found in ilamy.dev/demo : 
image

kcsujeet (approved):
Looks good. Thank you for the PR. 🙏

PR #176, opened 2026-05-28
fix: reset a reccuring event after editing width scope "all" 

### LINKED ISSUE #151: [BUG] Duplicating event after changing the whole series
On the demo move an event of "Daily standup" , choose "this event" on the scope dialog
In the past select on event of "Daily standup" move later , choose the "all" on the scope dialog
The previous element is duplicated

Image

maxdelorme:
I follow Google Agenda behaviour.
it fixes #151 

## What changed?
It ensure after a All editing event that exdates are empty and return only the parents

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [ ] Changes are tested


src/features/recurrence/utils/recurrence-handler.ts
```diff
@@ -325,15 +324,22 @@ export const updateRecurringEvent = ({
...
+			const parentUid = getEventParentUID(baseEvent)
+
 			const anchored = applyAllScopeUpdates(baseEvent, updates)
 			updatedEvents[baseEventIndex] = {
 				...baseEvent,
 				...updates,
 				start: anchored.start,
 				end: anchored.end,
 				rrule: anchored.rrule,
+				exdates: undefined,
```
kcsujeet, line 336: question: Clearing every EXDATE on scope "all" resurrects previously-deleted occurrences.

Edit-this (line ~263) and delete-this (line ~368) both write EXDATEs, but only edit-this also creates an override row. This change removes the override rows and also wipes all EXDATEs, so a delete-only EXDATE (an occurrence cancelled via "delete this", with no override) is cleared too, and that occurrence reappears after any "all" edit. Is that intended? If cancelled occurrences should survive an "all" edit, clear only the EXDATEs tied to the overrides being removed. No test covers delete-this-then-edit-all, so this path is currently unverified.

src/features/recurrence/utils/recurrence-handler.ts
```diff
@@ -325,15 +324,22 @@ export const updateRecurringEvent = ({
...
 			}
-			break
+
+			return updatedEvents.filter(
+				(e) =>
+					!(e.recurrenceId && !e.rrule && getEventParentUID(e) === parentUid)
```
kcsujeet, line 341: suggestion: Extract the detached-override check to a named predicate with named clauses.

`!(e.recurrenceId && !e.rrule && getEventParentUID(e) === parentUid)` is a three-clause condition under a negation, hard to parse at a glance. Pull it into a named function and name the two ideas inside it:

```ts
const isDetachedOverrideOfSeries = (e: CalendarEvent): boolean => {
  const isDetachedOverride = Boolean(e.recurrenceId) && !e.rrule
  const belongsToSeries = getEventParentUID(e) === parentUid
  return isDetachedOverride && belongsToSeries
}
return updatedEvents.filter((e) => !isDetachedOverrideOfSeries(e))
```

The filter then reads as "keep everything that is not a detached override of this series".

src/features/recurrence/utils/recurrence-handler.test.ts
```diff
@@ -117,6 +117,152 @@ describe('recurrence-handler utility tests', () => {
 			expect(result).toHaveLength(1)
 			expect(result[0].title).toBe('New Title')
 		})
+
+		it('should remove all detached overrides and keep only the base for scope all from a detached target', () => {
+			const baseEvent = createBaseRecurringEvent({
```
kcsujeet, line 122: suggestion: Extract test factories for the repeated daily series and detached override.

Across the new scope "all" tests here and in crud-operations.test.ts, the same daily-series base (same freq DAILY, dtstart, 2025-01-06 start/end) and the same `{ ...baseEvent, id, start, end, recurrenceId, rrule: undefined }` override spread are rebuilt five times, varying only id/exdates/dates. Two small local factories, e.g. `makeDailySeries(overrides)` and `makeOverride(base, overrides)`, would remove the boilerplate and make each test's actual variance obvious.

src/features/recurrence/utils/recurrence-handler.ts
maxdelorme: These is the Google Calendar behaviour.

But you ask a good question what should be the correct behaviour : 
if I changed the whole series by editing an event from the series : 
 - Is a detached override event should stay detached and override the updated series ?
 - is a deleted event should stay deleted ? 
 if I changed the whole series by editing an override event from the series : 
 - does the pop-up should not be shown ?
 - Is the detached override event should stay detached and override the series but the rest of the series move as the edited override event ?
 - is a deleted event should stay deleted ? 

Google behaviour is a "whole" editing reset the series and that's it. It resurrects previously deleted event and override events.

I you think this behaviour is not the correct one, please fill free to describe what you expect.

PR #181, opened 2026-05-31
refactor(plugins): introduce plugin runtime, route recurrence through it (phase 1)

kcsujeet:
Phase 1 of making recurrence a tree-shakable plugin. **Behavior-preserving refactor only** — no public API change, recurrence stays registered by default, nothing is tree-shakable yet (that is Phase 2).

## What this does

Introduces an internal plugin abstraction and routes the calendar engine, drag-and-drop, and event form through it, so Phase 2 can move recurrence to a subpath and drop it from the default bundle.

- **`src/lib/plugin.ts`** — the `IlamyPlugin` contract + `createPluginRuntime` (`expandEvents`, `getOwner`, `getFormSectionPlugins`).
- **`src/features/recurrence/recurrence-plugin.tsx`** — a built-in adapter implementing the contract by wrapping the existing recurrence functions/components.
- **Engine** — `getEventsForDateRange` expands via the runtime; scoped edit/delete delegate to the owning plugin.
- **dnd-context** — scope dialog rendered by the owning plugin (no static recurrence imports).
- **event-form** — recurrence editor + dialog rendered via plugins (capability-based form sections; ownership-based edit dialog).
- Providers thread an optional `plugins` prop (defaults to the recurrence plugin).

## Parity is the bar

All existing tests pass unchanged (896 total, including new contract/adapter tests). Two parity points handled explicitly:
- `ownsEvent` mirrors the old `isRecurringEvent` exactly (rrule, recurrenceId, or uid) so the scope dialog still shows for modified/generated instances.
- `expandEvent` receives the full event list so `generateRecurringEvents` can still merge detached overrides (regression test added).

## Still coupled on purpose (Phase 2 to-dos)

- `src/index.ts` still re-exports `generateRecurringEvents` / `isRecurringEvent` / `RRule` (public API unchanged here).
- The engine still imports `recurrencePlugin` as the default registration.
- `event-form.tsx` still imports `useRecurringEventActions` (dialog-state hook) — Phase 2 generalizes this out.

Phase 2 does the flip: move recurrence to `src/plugins/recurrence/`, drop the default, add the `@ilamy/calendar/plugins/recurrence` subpath export + `sideEffects:false`, the bundle gate, and fold in #176.

## Verification

- `bun run type-check` clean
- `bun test` — 896 pass / 0 fail
- `bun run build` succeeds


kcsujeet:
Superseded by the v2.0.0 plugin architecture. The contract was redesigned from the recurrence-shaped hooks (expandEvent / ownsEvent / renderEditDialog / renderFormSection) into a generic plugin API (transformEvents / managesEvent / renderSlot / contribute / views / provider), with recurrence becoming a true plugin built against the public API. That work continues on feat/v2-plugin-architecture, so closing this in its favor.

PR #191, opened 2026-06-11
fix(recurrence): edit recurring event should call correctly event han…

### LINKED ISSUE #162: [BUG] When editing a recurring event, event handler are not triggers correctly
- when editing a recurring event then choose scope **"this"** : update event handler is trigger: 
**expected** add and update with correct id and uid
- when editing a recurring event then choose scope **"this and following"** : update event handler is trigger
**expected** update with correct id and rrule update and add event trigger for the new series
- when editing a recurring event then choose scope **"all",** update event handler is trigger 
**expected** update with correct id and uid.
uid is necessary to remove all other override event in my storage that share this uid

maxdelorme:
## What changed?
fixes #162 

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

### DIFF
--- a/apps/demo/src/components/demo/demo-data.ts
+++ b/apps/demo/src/components/demo/demo-data.ts
@@ -18,14 +18,35 @@ export const handleDateClick = (info: CellInfo) => {
 
 export const handleEventAdd = (event: CalendarEvent) => {
 	alert(`Event added: ${event.title}`)
+	console.log('Added event:', {
+		id: event.id,
+		exdates: event.exdates,
+		recurrenceId: event.recurrenceId,
+		rrule: event.rrule,
+		uid: event.uid,
+	})
 }
 
 export const handleEventUpdate = (event: CalendarEvent) => {
 	alert(`Event updated: ${event.title}`)
+	console.log('Updated event:', {
+		id: event.id,
+		exdates: event.exdates,
+		recurrenceId: event.recurrenceId,
+		rrule: event.rrule,
+		uid: event.uid,
+	})
 }
 
 export const handleEventDelete = (event: CalendarEvent) => {
 	alert(`Event deleted: ${event.title}`)
+	console.log('Deleted event:', {
+		id: event.id,
+		exdates: event.exdates,
+		recurrenceId: event.recurrenceId,
+		rrule: event.rrule,
+		uid: event.uid,
+	})
 }
 
 // Demo resources
--- a/docs/logs/2026-06-09.md
+++ b/docs/logs/2026-06-09.md
@@ -2,13 +2,25 @@
 
 ## Changes
 
-- **recurrence**: Fixed "this" scope delete leaving detached overrides visible after editing then deleting a modified occurrence.
+- **[recurrence / callbacks]**: Fixed duplicate firing of `onEventUpdate`/`onEventAdd`/`onEventDelete` for scoped recurring edits in dev.
+
+  **Root cause (two layers):**
+  1. **StrictMode double-invocation** — `useScopedEventMutation.handleConfirm` ran `applyScopedEdit`/`applyScopedDelete` inside a `setState` updater. React StrictMode (enabled in the demo via `frontend.tsx`) intentionally runs updaters twice in dev, so every scoped confirm fired persistence callbacks twice.
+  2. **Wrong callback payload** — `applyScopedEdit` always called `onEventUpdate` with a synthetic merged UI instance (`{ ...event, ...updates }`, e.g. `recurring-1_2`) before applying the plugin mutation, instead of emitting the real stored rows.
+
+  **Fix:** Move side effects out of the setState updater (ref + apply outside). Route recurring edits through `RecurringUpdateResult` / `PluginMutationResult` so the engine dispatches `updated` → `onEventUpdate` and `added` → `onEventAdd` with persisted event ids.
 
 ## Files Modified
 
-- `packages/recurrence/src/utils/recurrence-handler.ts` - Use `recurrenceId` for occurrence identity on delete; remove matching detached overrides; dedupe EXDATEs.
-- `packages/recurrence/src/utils/crud-operations.test.ts` - Regression test for edit-then-delete on modified override.
+- `packages/calendar/src/hooks/use-scoped-event-mutation.ts` - apply mutations outside setState updater
+- `packages/calendar/src/hooks/use-scoped-event-mutation.test.ts` - StrictMode regression test
+- `packages/calendar/src/hooks/use-calendar-engine.ts` - `PluginMutationResult` dispatch in `applyScopedEdit`
+- `packages/recurrence/src/utils/recurrence-handler.ts` - `RecurringUpdateResult` from `updateRecurringEvent`
+- `packages/types/src/index.ts` - `PluginMutationResult` on plugin contract
+- Tests: engine, provider, crud-operations updated
 
 ## Notes
 
-Deleting a modified instance ("this" scope) now removes the detached override from stored events and keys EXDATE by original occurrence (`recurrenceId`), not the override's moved start time.
+- Scope `"this"` still emits one `onEventUpdate` (base + EXDATE) and one `onEventAdd` (override) — that is correct RFC behaviour, not a duplicate.
+- Plain (non-recurring) add/update/delete was never doubled; only scoped mutations through the recurrence scope dialog were affected.
+- After changing library source locally, run `bun run build:lib` and restart `bun dev` so the demo picks up fresh `dist/`.
--- a/packages/calendar/src/hooks/use-calendar-engine.test.ts
+++ b/packages/calendar/src/hooks/use-calendar-engine.test.ts
@@ -565,12 +565,14 @@ describe('useCalendarEngine', () => {
 	describe('recurring events', () => {
 		it('should update recurring event with scope all', () => {
 			const onEventUpdate = vi.fn()
+			const onEventAdd = vi.fn()
 			const events = [createRecurringEvent()]
 			const { result } = renderHook(() =>
 				useCalendarEngine({
 					...defaultConfig,
 					events,
 					onEventUpdate,
+					onEventAdd,
 					plugins: [recurrencePlugin()],
 				})
 			)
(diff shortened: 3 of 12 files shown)


packages/recurrence/src/utils/recurrence-handler.ts
```diff
@@ -273,36 +273,85 @@ interface UpdateRecurringEventProps {
...
+	added: CalendarEvent[]
+}
+
 export const updateRecurringEvent = ({
 	targetEvent,
 	updates,
 	currentEvents,
 	scope,
-}: UpdateRecurringEventProps): CalendarEvent[] => {
+}: UpdateRecurringEventProps): RecurringUpdateResult => {
```
kcsujeet, line 290: issue (blocking): `recurrence-handler.test.ts` was not migrated to the new return shape, so CI is red.

`updateRecurringEvent` now returns `RecurringUpdateResult`, but `src/utils/recurrence-handler.test.ts:110-119` still asserts on it as an array (`toHaveLength(1)`, `result[0].title`). Type-check fails with TS7053 in the Build, Type-check and Test job. Either migrate that test to `.events` like `crud-operations.test.ts`, or reuse the same wrapper helper.

packages/recurrence/src/utils/recurrence-handler.ts
```diff
@@ -273,36 +273,85 @@ interface UpdateRecurringEventProps {
...
+				if (overrideIndex === -1) {
+					throw new Error('Detached override not found')
+				}
+				updatedEvents[overrideIndex] = detachedOverride
+				return {
+					events: updatedEvents,
+					updated: [detachedOverride],
+					added: [],
+				}
+			}
```
kcsujeet, line 341: issue: the stored-override branch mutates the base row without reporting it, and has no test.

Lines 313-317 rewrite the base (uid backfill, EXDATE dedupe) before this branch, but `updated` only contains the override, so a base row that just gained a uid is never emitted through `onEventUpdate` and consumer storage desyncs. Every scope-this test targets a generated instance; nothing covers this in-place path or its line 333 throw. Either leave the base untouched here or include it in `updated` when it changed, and add a test that edits a stored override.

packages/recurrence/src/utils/recurrence-handler.ts
```diff
@@ -273,36 +273,85 @@ interface UpdateRecurringEventProps {
...
+					? targetEvent.recurrenceId
+					: targetEvent.start.toISOString()
 
-			// Create standalone modified event with recurrenceId
-			const modifiedEventId = `${targetEvent.id}_modified_${Date.now()}`
-			const modifiedEvent: CalendarEvent = {
+			const existingExdates = baseEvent.exdates || []
+			const nextExdates = existingExdates.includes(recurrenceId)
+				? existingExdates
+				: [...existingExdates, recurrenceId]
```
kcsujeet, line 311: nitpick: two multi-line ternaries.

`recurrenceId` reads cleaner as an if-assignment, and `nextExdates` as an early `includes` guard, per the project's flat-over-wrapped-ternary convention.

packages/recurrence/src/utils/recurrence-handler.ts
```diff
@@ -342,38 +391,46 @@ export const updateRecurringEvent = ({
...
-			const isDetachedOverrideOfSeries = (e: CalendarEvent): boolean => {
-				const isDetachedOverride = Boolean(e.recurrenceId) && !e.rrule
-				const belongsToSeries = getEventParentUID(e) === parentUid
-				return isDetachedOverride && belongsToSeries
+			updatedEvents[baseEventIndex] = updatedBaseEvent
+
+			const events = updatedEvents.filter(
+				(e) =>
+					!(e.recurrenceId && !e.rrule && getEventParentUID(e) === parentUid)
+			)
```
kcsujeet, line 421: suggestion: restore the single uid const and the named predicate.

`parentUid` and `seriesUid` are the same `getEventParentUID(baseEvent)` call twice. The filter also inlines three clauses that the previous revision had as the named `isDetachedOverrideOfSeries` plus a comment explaining the Google-Calendar reset behavior; the name and the comment carried the WHY and should survive this refactor.

packages/types/src/index.ts
```diff
@@ -107,6 +107,15 @@ export interface PluginMutationArgs {
 	scope: unknown
 }
 
+/** Structured result from `applyEdit` when a mutation updates multiple stored rows. */
+export interface PluginMutationResult {
```
kcsujeet, line 111: suggestion: reuse `PluginMutationResult` instead of duplicating it as `RecurringUpdateResult`.

The two interfaces are field-for-field identical, so they can drift silently. Export `PluginMutationResult` from `@ilamy/calendar`'s public index (next to `PluginMutationArgs`, `src/index.ts:22`) and alias or use it directly in the recurrence package.

packages/calendar/src/hooks/use-calendar-engine.ts
```diff
@@ -86,6 +90,15 @@ const VIEW_UNITS: Record<string, ManipulateType> = {
...
 }
 
+const isPluginMutationResult = (
+	result: CalendarEvent[] | PluginMutationResult
+): result is PluginMutationResult =>
+	typeof result === 'object' &&
+	result !== null &&
+	'events' in result &&
+	'updated' in result &&
+	'added' in result
```
kcsujeet, line 100: suggestion (if-minor): `!Array.isArray(result)` discriminates this union in one clause.

The input type is exactly `CalendarEvent[] | PluginMutationResult`, and `Array.isArray` is declared as a type predicate (`arg is any[]`, lib.es5.d.ts), so TypeScript narrows the false branch to `PluginMutationResult` without the three `in` checks.

packages/calendar/src/hooks/use-calendar-engine.ts
```diff
@@ -262,12 +275,26 @@ export const useCalendarEngine = (
...
 				return
 			}
+			const editResult = manager.applyEdit({
+				event,
+				updates,
+				currentEvents,
+				scope,
+			})
+			if (isPluginMutationResult(editResult)) {
+				for (const storedEvent of editResult.updated) {
```
kcsujeet, line 285: thought: scoped delete has the same stale-payload shape this PR fixes for edit.

`applyScopedDelete` still emits the UI instance through `onEventDelete`, and a scope-this delete never reports the base row's new EXDATE through `onEventUpdate`. Worth a follow-up issue extending `PluginMutationResult` handling to `applyDelete` so both paths agree.

packages/calendar/src/hooks/use-scoped-event-mutation.ts
```diff
@@ -37,6 +37,10 @@ export function useScopedEventMutation(
 	)
 	const [dialogState, setDialogState] =
 		useState<ScopedMutationDialogState>(CLOSED)
+	const dialogStateRef = useRef(dialogState)
+	useEffect(() => {
+		dialogStateRef.current = dialogState
+	}, [dialogState])
```
kcsujeet, line 43: suggestion: read `dialogState` directly instead of mirroring it into a ref.

Moving the side effects out of the updater is the right fix, but the ref plus sync effect can be replaced by reading `dialogState` in `handleConfirm` and adding it to the deps. The dialog re-renders whenever `dialogState` changes, so every render gets a fresh callback.

packages/calendar/src/hooks/use-calendar-engine.test.ts
```diff
@@ -584,9 +586,161 @@ describe('useCalendarEngine', () => {
...
 
 			expect(onEventUpdate).toHaveBeenCalledTimes(1)
+			expect(onEventAdd).not.toHaveBeenCalled()
 			const updatedEvent = onEventUpdate.mock.calls[0][0]
 			expect(updatedEvent.title).toBe('Updated Meeting')
 			expect(updatedEvent.id).toBe('recurring-1')
+			expect(updatedEvent.uid).toBe('recurring-1@ilamy.calendar')
+		})
+
+		it('should emit onEventUpdate and onEventAdd with correct ids for scope this', () => {
```
kcsujeet, line 596: nitpick: new tests copy the renderHook block and use bracket indexing.

`code-style.md` asks for a local `render<Name>Hook` helper instead of copy-pasting the `renderHook` call (five new copies here), and `.at()` over `[0]` (`mock.calls[0][0]`, `rawEvents[0]` here, `updated[0]` in `crud-operations.test.ts`).

apps/demo/src/components/demo/demo-data.ts
```diff
@@ -18,14 +18,35 @@ export const handleDateClick = (info: CellInfo) => {
 
 export const handleEventAdd = (event: CalendarEvent) => {
 	alert(`Event added: ${event.title}`)
+	console.log('Added event:', {
```
kcsujeet, line 21: nitpick: debug console.log dumps in the demo handlers.

If these were for verifying the fix, drop them; the alerts already signal the callbacks. If they are meant to stay as demo output, one log of the whole event says the same with less code.

docs/logs/2026-06-09.md
```diff
@@ -2,13 +2,25 @@
 
 ## Changes
 
-- **recurrence**: Fixed "this" scope delete leaving detached overrides visible after editing then deleting a modified occurrence.
+- **[recurrence / callbacks]**: Fixed duplicate firing of `onEventUpdate`/`onEventAdd`/`onEventDelete` for scoped recurring edits in dev.
```
kcsujeet, line 5: todo: this rewrites the merged June 9 log entry instead of adding a new one.

The replaced entry documented the #189 override-delete fix that is already on main. Restore it and put this PR's notes in a log file dated for this work; the log rules append per day, they do not revise history.

kcsujeet:
Superseded by #194. The v2 architecture merge (#192) conflicted this branch, so we re-implemented the fix on the new structure and credited you as co-author. The new PR also collapses the duplicate result type, fixes the stored-override branch so the base row's EXDATE reaches `onEventUpdate`, and migrates both recurrence test suites (which is what left CI red here). Thanks for the original work and the clear issue repro.

Issue #156, opened 2026-05-19
[BUG] When deleting an override event of recurrent serie, an update event is trigger for nothing


maxdelorme:
- 🐞  When deleting a override  event it fired a deleted event with the id <previous_id>_modified....  and an update event
**Expected** only on deleted event , the updated event should not be trigger
- ✅ When deleting directly one event of a recurring event directly choosing scope **this** it fired an updated event with base ID
  Expected : update event should be fired with the base Id with a exdates updated with a the date of the deleting event
- ✅ When deleting an recurring event directly choosing scope **this and following events** it fired a deleted event with the id <previous_id>_0
 Expected : update event should be fired with the base Id with a rrule.until updated with a the date of the deleting event
- ✅ When deleting an recurring event directly choosing scope **All** event it fired a deleted event with the id <previous_id>_0
 Expected : delete event should be fired with the base Id

kcsujeet:
Fixed in **v2.0.0** (#199). Deleting an override of a recurring series no longer fires a spurious update event.


Note: v2.0.0 is a major release with breaking changes. Please read the [v2 migration guide](https://ilamy.dev/docs/help/migration-to-v2/) before upgrading.

PR #194, opened 2026-06-12
fix(recurrence)!: scoped recurring edits fire onEventUpdate/onEventAdd with real stored rows

### LINKED ISSUE #162: [BUG] When editing a recurring event, event handler are not triggers correctly

kcsujeet:
Closes #162. Re-implements #191 on the v2 architecture (its branch conflicted after the v2 merge landed on main).

## Problem

Editing a recurring event through the scope dialog fired the wrong persistence callbacks: `applyScopedEdit` emitted a synthetic merged UI instance (e.g. `recurring-1_2`) via `onEventUpdate` and never fired `onEventAdd` for the new detached override or split series. Under React StrictMode the callbacks also fired twice.

## Fix

Scoped recurring edits now emit the real stored rows:

- **scope "this"** — `onEventUpdate` for the base row (with its new EXDATE, keeping the series id and uid) + `onEventAdd` for the detached override (carrying the series uid).
- **scope "following"** — `onEventUpdate` for the terminated original series + `onEventAdd` for the new split series.
- **scope "all"** — single `onEventUpdate` for the base row (correct id + uid), no `onEventAdd`.
- No double-firing under StrictMode (the mutation no longer runs inside a `setState` updater).

`updateRecurringEvent` now returns a structured `PluginMutationResult { events, updated, added }`, and the data slice dispatches `updated → onEventUpdate`, `added → onEventAdd`.

## Improvements over #191

- Collapses #191's two identical result types (`RecurringUpdateResult` + `PluginMutationResult`) into one published `PluginMutationResult`.
- Fixes the stored-override branch so the base row's dedup'd EXDATE/uid reaches `onEventUpdate` (#191 dropped it).
- Migrates both recurrence test suites to the structured return — #191 left `recurrence-handler.test.ts` unmigrated, which was its red CI.

## Test plan

- [x] `bun run ci` green: 158 (recurrence) + 783 (calendar), 0 fail
- [x] Fallow audit green (`--changed-since main`)
- [x] Failing-first tests for each scope's callback ids/uids + a StrictMode single-fire test
- [x] New test pinning the stored-override branch reporting the base row in `updated`

Credits @maxdelorme (co-authored) for the original #191 implementation and the issue repro.


maxdelorme:
I kcsujeet, thank for updating my code.

I tested your branch , these are the issues have found so far: 
- 🐞 When editing an overriding event then choosing again scope **this** : 
   I see 2 update events are fired one for the override event, one for the base/parent one (with the same event than previously)
  expected : only one update event should be fired: the one of the override event
- 🐞   When editing an event of the series (I would say on day D)  **before** override events E, F  and choosing scope **this event and following** : the overrides E, F events wich follow the currently editing are unchanged. 
 expected : the following override events E,F should disappear (deleting event fired with an array of id ? ) , the updating/base/parent event should have rrule with `until` day D and `exdates` without the `recurrenceId` of E and F 
- ✅ When deleting an override event it fired a delete event correctly 
- 🐞 When deleting an recurring event directly choosing scope **this** it fired a deleted event with the `id` `<previous_id>_0`
Expected : update event should be fired with the base Id with a `exdates` updated with a the date of the deleting event 
- 🐞 When deleting an recurring event directly choosing scope **this and following** it fired a deleted event with the `id` `<previous_id>_0`
Expected : update event should be fired with the base Id with a `rrule.until` updated with a the date of the deleting event 
- 🐞 When deleting an recurring event directly choosing scope **All event** it fired a deleted event with the `id` `<previous_id>_0`
Expected : delete event should be fired with the base Id

Note : the last 3 items are part of #156.

Thank you for this amazing project !!!

kcsujeet:
@maxdelorme both issues from your testing are fixed and this is now merged to main.

- Edit scope **"this"** on a stored override: fires a single `onEventUpdate` for the override only. The base row is now reported only when its EXDATE/uid actually changed, so the redundant second update is gone.
- Edit **"this and following"** before later overrides: the original series is terminated with `until`, and the overrides whose occurrence falls after the split are reported via `onEventDelete` (cascaded out) instead of being left orphaned.

Both are covered by new tests. Credited you as co-author on the commits for the original #191 work and the repro.

Could you pull `main` and verify on your end? If you spot anything else, drop it here and I'll get it fixed.

PR #199, opened 2026-06-14
fix: deleting an overriding event should not triggerd an update event

### LINKED ISSUE #156: [BUG] When deleting an override event of recurrent serie, an update event is trigger for nothing

maxdelorme:
## What changed?

fix #156 

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [ ] Changes are tested

### DIFF
--- a/docs/logs/2026-06-14.md
+++ b/docs/logs/2026-06-14.md
@@ -1,0 +1,17 @@
+# Development Log - 2026-06-14
+
+## Changes
+
+- **[recurrence handler — delete "this"]**: Scope-"this" delete now emits exactly one persistence callback. When a stored override is dropped, only `deleted` is reported; otherwise only `updated` (new EXDATE on the base). In-memory state still always applies the EXDATE and removes the override row so the occurrence stays excluded even when the callback is delete-only.
+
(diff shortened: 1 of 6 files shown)


kcsujeet (approved):
Looks good. Thanks for the PR.

PR #211, opened 2026-06-17
feat(calendar): add getCellClassName prop for customizable cell styling

### LINKED ISSUE #210: [FEATURE]: add an helper to customize cell className
I would like to displays specific period on unavailability. during this period adding event is allowed (is not a disabled period)
For that I would like a new prop on IlamyCalendarProps : getCellClassName

maxdelorme:
fixes #210 

Introduced a new `getCellClassName` prop in `IlamyCalendarProps` to allow for visual-only cell styling, enabling cells to remain clickable while applying custom classes. Updated relevant components and context to support this feature, along with unit and integration tests to ensure functionality.

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [ ] Changes are tested

### DIFF
--- a/packages/calendar/src/features/calendar/contexts/calendar-context/provider.tsx
+++ b/packages/calendar/src/features/calendar/contexts/calendar-context/provider.tsx
@@ -33,6 +33,7 @@ export interface CalendarProviderProps {
 	onEventClick?: (event: CalendarEvent) => void
 	onCellClick?: (info: CellInfo) => void
 	isCellDisabled?: (info: CellInfo) => boolean
+	getCellClassName?: (info: CellInfo) => string
 	onViewChange?: (view: CalendarView) => void
 	onEventAdd?: (event: CalendarEvent) => void
 	onEventUpdate?: (event: CalendarEvent) => void
@@ -98,6 +99,7 @@ const useCalendarContextValue = (
 		onEventClick,
 		onCellClick,
 		isCellDisabled,
+		getCellClassName,
 		onViewChange,
 		onEventAdd,
 		onEventUpdate,
@@ -174,6 +176,7 @@ const useCalendarContextValue = (
 			onEventClick: handleEventClick,
 			onCellClick: handleDateClick,
 			isCellDisabled,
+			getCellClassName,
 			locale,
 			timezone,
 			disableCellClick,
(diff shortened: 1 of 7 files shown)


packages/calendar/src/features/calendar/types/index.ts
```diff
@@ -153,6 +153,12 @@ export interface IlamyCalendarProps {
 	 * business hours or this predicate disables it).
 	 */
 	isCellDisabled?: (info: CellInfo) => boolean
+	/**
+	 * Returns extra CSS classes for a cell based on custom logic (unavailability
+	 * styling, on-call windows, etc.). Unlike `isCellDisabled`, this is purely
+	 * visual — cells remain clickable and accept drag-drops.
+	 */
+	getCellClassName?: (info: CellInfo) => string | undefined
```
kcsujeet: suggestion: Return `string` rather than `string | undefined`.

`cn()`/clsx drops `''` exactly like `undefined` (verified: `cn('base', '', 'extra')` is `'base extra'`), so the `| undefined` adds no behavior. It also diverges from the sibling `isCellDisabled?: (info) => boolean` at line 155, which returns a definite type. Tightening to `=> string` (return `''` for no class) keeps the two cell callbacks consistent.

packages/calendar/src/features/calendar/contexts/calendar-context/provider.tsx
```diff
@@ -33,6 +33,7 @@ export interface CalendarProviderProps {
 	onEventClick?: (event: CalendarEvent) => void
 	onCellClick?: (info: CellInfo) => void
 	isCellDisabled?: (info: CellInfo) => boolean
+	getCellClassName?: (info: CellInfo) => string | undefined
```
kcsujeet: **suggestion:** 

this should be only `getCellClassName?: (info: CellInfo) => string`

mdelorme-mediway: done

kcsujeet (approved):

kcsujeet:
Thank you for the PR 🙏

PR #213, opened 2026-06-18
fix: resource picker missing on new event for calendar Resource

### LINKED ISSUE #212: [BUG] invisible created event in resource Calendar
**Steps to reproduce**
- When creating an event in resource Calendar with the New Button 
the modal with the event form appears
- when validating the modal shows off and I can't see the event because no resource is associated with the newly created event

**Expecting**
the event form has a resource picker when click on the + button not in an empty cell of the resouceCalendar

maxdelorme:
## What changed?
it fixes #212 

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested


packages/calendar/src/features/calendar/components/event-form/event-form.tsx
```diff
@@ -207,9 +227,25 @@ export const EventForm: React.FC<EventFormProps> = ({
 		}
 	}, [isAllDay])
 
+	useEffect(() => {
+		setSelectedResourceId(initialSelectedResourceId)
+	}, [initialSelectedResourceId])
```
kcsujeet: suggestion: Drop this effect; it duplicates the `useState` initializer.

`useState(initialSelectedResourceId)` already seeds the state on mount. None of the form's other fields (title, dates, color, allDay) sync to `selectedEvent` via an effect, so the form is initialized fresh per open. This effect only does something if the form is reused for a different event without remounting, which nothing else here supports.

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
```diff
@@ -287,6 +324,11 @@ export const EventForm: React.FC<EventFormProps> = ({
 		['endTime', 'end-time', endTime, handleEndTimeChange, endConstraints],
 	] as const
 
+	let resourceSelectValue: string | undefined
+	if (selectedResourceId !== undefined) {
+		resourceSelectValue = String(selectedResourceId)
+	}
```
kcsujeet: suggestion: Collapse to optional chaining.

`const resourceSelectValue = selectedResourceId?.toString()` gives the same `string | undefined` (both `number` and `string` have `.toString()`, and `undefined?.toString()` is `undefined`). The `let`/`if`/`String()` guard only exists to avoid `String(undefined)` producing the literal `"undefined"`.

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
```diff
@@ -126,7 +135,18 @@ export const EventForm: React.FC<EventFormProps> = ({
 		location: initialLocation = '',
 	} = selectedEvent ?? {}
 
-	const effectiveBusinessHours = useEffectiveBusinessHours(resourceId)
+	const hasResources = resources.length > 0
+	const initialSelectedResourceId = resourceId ?? resourceIds?.[0]
```
kcsujeet: todo: Use `.at(0)` instead of `[0]`.

Project convention (`.agents/rules/code-style.md`) is `resourceIds?.at(0)` over `resourceIds?.[0]`.

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
```diff
@@ -218,7 +254,7 @@ export const EventForm: React.FC<EventFormProps> = ({
 			title: formValues.title,
 			start: startDateTime,
 			end: endDateTime,
-			resourceId,
+			resourceId: effectiveResourceId,
```
kcsujeet: note: Editing a cross-resource event collapses it to a single resource.

For an event with `resourceIds`, the submit payload writes `resourceId: effectiveResourceId` (which resolves to `resourceIds[0]`) and never carries `resourceIds`, so the multi-resource assignment is lost on save (same on the update payload at line 270). The form has always been single-resource, so this is pre-existing. Flagging it because this PR now actively writes that `resourceId`.

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
mdelorme-mediway: done

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
mdelorme-mediway: done

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
mdelorme-mediway: done

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
mdelorme-mediway: Best to create a new bug for this , don't you think ?
`Manage in calendar resources multi-resources events`

PR #218, opened 2026-06-23
refactor(event-form): migrate to react-hook-form + valibot with composed field components

kcsujeet:
Migrates the event form from hand-rolled `useState`/validation to **react-hook-form + valibot**, with a reusable field kit and a decomposed, context-driven structure. Opening this as an evaluation: the open question is whether the consumer bundle cost is worth the cleanup.

## What

- **react-hook-form + valibot** (`@hookform/resolvers/valibot`): one `useForm` + a valibot schema replace ~10 `useState` hooks and the manual `onChange`/validation. Validation errors now actually render (previously a failed check silently blocked submit).
- **`@ilamy/ui` Field**: vendored a focused subset of shadcn's `Field` (`Field`, `FieldLabel`, `FieldError`) for accessible label/error markup.
- **`components/form/*`**: form-agnostic field components (`FormInput`, `FormSelect`, `FormCheckbox`, `FormDatePicker`, `FormTimePicker`, `FormColorPicker`) + a shared `FieldWrapper`. Each reads `control` from `FormProvider` context via `useController` — no `control` prop, no form-type generic at call sites.
- **Decomposition**: the form is split into logical sections (`EventDateTimeFields`, `EventResourceField`, `EventColorField`, `EventFormFooter`) that pull state through `useFormContext`/`useWatch` instead of props. `event-form.tsx` is now a thin composition root (499 → ~150 lines).
- **Dedup**: `buildDateTime`/`buildEndDateTime` collapsed into one pure `buildDateTime(date, time)`; submit mapping extracted to `toEventFields` + `persistEvent`.

## Bundle impact

- **Published `@ilamy/calendar` dist: +1.3 KB gz** (own code + vendored Field). The three form libs are external `dependencies` (bunup externalizes deps/peerDeps by default), so they are NOT bundled into dist.
- **Consumer bundle: ~+14 KB gz** (the external libs land in the consumer's build). Demo build: 274.96 → 289.04 KB gz.

## Tests

- Behavior preserved: the existing event-form integration spec stays green (808 calendar tests, + plugins/recurrence/agenda). The 3 removed util unit tests (all-day branch) were relocated into the all-day submission integration test.
- `bun run type-check`, biome, and the Fallow gate (`--gate new-only`) all pass.

## Open question for review

Is ~14 KB gz on the consumer bundle worth the cleaner form, the reusable `components/form/*` kit, and rendered validation errors? If the event form stays the only form, that's a lot for one form; if more forms/validation are coming, the kit amortizes.


kcsujeet:
Closing as stale. This was opened in June as an evaluation rather than a change to merge, and the question it asked has effectively been answered by three months of not merging it: ~14 KB gz on every consumer's bundle is a lot to pay for an internal cleanup of one form, however much nicer the result reads.

Nothing here is lost. The branch `refactor/event-form-rhf` stays, and the measurements are in the description above: +1.3 KB gz in the published dist, ~14 KB gz in a consumer build (demo 274.96 → 289.04 KB gz), with the three form libraries staying external.

Worth separating out if anyone picks this up again: the decomposition of `event-form.tsx` from 499 lines into sections, and the `buildDateTime` dedup, cost no bundle at all and stand on their own. The validation-errors-actually-render fix is a real bug fix independent of react-hook-form. Those could land without the dependency; the form-library migration is the part that has to justify the 14 KB.

PR #230, opened 2026-07-03
feat(calendar): add onMoreEventsClick to override built-in "+N more" dialog

sangeetha-armtek:
## Summary

Adds an optional `onMoreEventsClick` prop to `IlamyCalendar`. When provided, clicking the "+N more" overflow indicator in a grid cell invokes the callback with the cell's day and its full list of events, instead of opening the built-in all-events dialog. When omitted, the default dialog opens — fully backward compatible.

## Motivation

The built-in overflow dialog was hardcoded, so consumers had no way to render their own popover/drawer or navigate (e.g. jump to day view) when a user clicks "+N more". This hands that interaction back to the consumer, in line with the library's pluggable-UI direction.

## API

```ts
onMoreEventsClick?: (day: Dayjs, events: CalendarEvent[]) => void
```

- `day` — the cell's day
- `events` — all events for that day (the full list, matching what the built-in dialog shows)

## Changes

- Thread the prop through `IlamyCalendarProps` → `CalendarProvider` → calendar context (both `useMemo` dependency arrays updated)
- `grid-cell`: `showAllEvents` defers to the callback when set, otherwise opens the built-in dialog
- Works for both the regular and resource calendars (unified under one `CalendarProvider`)
- Docs: added the prop to `types-and-interfaces.md`
- Playground: added a demo handler wired through `SharedCalendarProps` into both calendar variants

## Testing

- New `grid-cell-more-events.test.tsx` — 3 tests: indicator renders, default-dialog branch, callback-override branch (verifies args and that the dialog is bypassed)
- Full suite green (calendar 838, recurrence 165, drag-to-create 56, agenda 19); type-check clean for `@ilamy/calendar` and the playground


packages/calendar/src/components/grid-cell-more-events.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+		<CalendarProvider
+			dayMaxEvents={1}
+			events={mockEvents}
+			initialDate={initialDate}
+			onMoreEventsClick={onMoreEventsClick}
+		>
+			<GridCell day={initialDate} shouldRenderEvents={true} />
+		</CalendarProvider>
+	)
+}
```
kcsujeet: suggestion (non-blocking): This file re-implements the setup harness already in the sibling `grid-cell-spacing.test.tsx`: same `initialDate`, near-identical `mockEvents`, a `renderGridCell` helper that only varies one `CalendarProvider` prop, and the same `beforeEach(cleanup)`.

Consider generalizing one shared `renderGridCell(overrides)` helper (accepting `dayMaxEvents`, `eventSpacing`, `onMoreEventsClick`, etc.) that both suites reuse, or folding these three tests into `grid-cell-spacing.test.tsx`, so the GridCell setup lives in one place.

sangeetha-armtek: Done — folded the three tests into `grid-cell-spacing.test.tsx` behind a shared `renderGridCell(options)` helper (now takes `dayMaxEvents`, `eventSpacing`, `eventHeight`, `events`, `onMoreEventsClick`), and added a `mkEvent(id)` factory to drop the duplicated event literals. Removed the standalone `grid-cell-more-events.test.tsx`. Thanks for the suggestion!

kcsujeet:
Thanks for the PR. This will go out in next version.

kcsujeet:
Shipped in v2.0.1 — https://github.com/kcsujeet/ilamy-calendar/releases/tag/v2.0.1

Thanks for the contribution!

PR #241, opened 2026-07-25
perf: fix year view 516-query defect and event/date layer constants

mattanderson-io:
Closes #245

## What changed?

Removes a constant-factor bottleneck in the event/date layer and fixes a year-view defect that made a single recurring event cost hundreds of milliseconds per render.

The algorithms here were never the problem — every hot path measured is already linear or linearithmic in the event count. The cost was in the date layer, and in three places that re-scanned the whole event array once per day.

**Headline finding: everything is dramatically worse when the `timezone` prop is set.** `startOf('day')` costs ~48µs on a `.tz()` instance versus ~350ns on a plain one, and the per-day loops called it inside the event loop. At 512 events, building one week row's day map took **2.9 seconds** with a timezone configured.

Each package's `testing-library.ts` clears the default timezone, so a suite runs timezone-free unless it opts in. Three suites do opt in — `date-utils.test.ts` (21 `setDefault` calls), `use-calendar-engine.test.ts` (2) and `recurrence-timezone.test.ts` (2) — but none covered the per-day map, agenda grouping, or the year-view grid, which is where the cliff lives. That gap is closed here.

### Measured, Apple M4 / Bun 1.3.14

| Kernel | Load | tz cleared | tz set |
|---|---|---|---|
| `eventOverlapsRange` full scan | 512 events | 6 240 → 37 ns/event (168x) | ~920x |
| Week day-map, one 7-day row | 512 events | 16.5 ms → 18.4 µs (**896x**) | 2 901 ms → 44.9 µs (**64 651x**) |
| Agenda grouping, 30-day window | 512 events | 31.1 ms → 0.45 ms (69x) | 4 476 ms → 4.8 ms (**938x**) |
| Year view | 1 recurring event | 316 → 12.4 ms (25x) | 1 299 → 389 ms |
| Year view | 16 recurring events | 5 444 → 152 ms (36x) | 17 821 → 4 637 ms |
| `startOf('day')` | — | unchanged | 81 → 48 µs (1.6x) |

A 365-day agenda window with 2 048 events was dropped from the benchmarks: the current implementation with a timezone set does not terminate in reasonable time.

### Commits

1. `perf:` compare event ranges as epoch millis — new `@ilamy/utils/events` with a `WeakMap` bounds cache
2. `perf:` shared `buildDayIndex` / `bucketEventsByDay` helpers
3. `perf:` build the week day-map in one bucketing pass
4. `fix:` derive year view counts from one query instead of 516
5. `perf:` group agenda events in one pass
[...]

PR #242, opened 2026-07-25
fix: date picker throws when a recurrence end date is typed

mattanderson-io:
## What changed?

Typing or picking an end date for a recurring event throws:

```
RangeError: invalid time zone: YYYY-MM-DD
```

`DatePicker` parses typed input with `dayjs(raw, DATE_INPUT_FORMAT)`. The configured dayjs in `@ilamy/utils/dayjs` wraps the constructor and forwards every argument to `dayjs.tz(...)`, whose signature is `tz(input, timezone)` — so the format string is read as a timezone name.

The component backs the "until" date in `recurrence-end-fields.tsx`, so this fires on every keystroke in the recurrence editor.

### Fix

`raw` comes from a native `<input type="date">` and is already an ISO date string, so the single-argument constructor parses it correctly and still honours the configured timezone. `DATE_INPUT_FORMAT` stays for display formatting, with a comment noting why it must not be passed as a parse format.

### Tests

`date-picker.test.tsx` — the component had none. Covers typed input, clearing the field, unparseable input, a leap day, and display formatting. **Two of the five fail before this change** with the RangeError above.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

---

Found while performance-profiling the date layer (#241), but this is independent of that work and branched off `main`.

Worth noting the underlying sharp edge: because the wrapper forwards extra arguments to `dayjs.tz()`, **any** `dayjs(input, format)` call in this codebase throws. This was the only such call site. A more general fix — having the wrapper distinguish a parse format from a timezone — would prevent the next occurrence, but that changes wrapper semantics and seemed out of scope for a bug fix.


kcsujeet:
Shipped in v2.0.2: https://github.com/kcsujeet/ilamy-calendar/releases/tag/v2.0.2

Thanks for the fix. Followed up in #243 so a parse-format argument is now a compile error.

kcsujeet:
Quick update: you probably want v2.0.3 rather than v2.0.2. The view headers were animating the whole header cell on every navigation, so I had to cut another release: https://github.com/kcsujeet/ilamy-calendar/releases/tag/v2.0.3

Your fix is in both.

PR #250, opened 2026-08-13
feat(events)!: treat event `end` as exclusive, and stop the form dropping seconds

### LINKED ISSUE #248: [Bug] Multi-day event spans are wrong: form drops sub-minute precision, and date grids read `end` as inclusive
Two separate defects make multi-day and all-day events land on the wrong days. They were found together while investigating #247 (which is only about ISO offsets being discarded during parsing, and is fixed separately). Part A is a plain bug with a safe fix. Part B is a convention mismatch whose fix is a breaking change.

The goal for both: follow **RFC 5545** and **Google Calendar**.

---

### Symptom

In the demo, three events were created the same way, each from one day at 12:00 AM to the next day at 12:00 AM, with `timezone` set to `America/Vancouver`:

- May 11 to May 12: renders as a **two-day** bar
- May 19 to May 20: renders as a **two-day** bar
- May 13 to May 14: renders as a **one-day** bar

Same shape, different result, and the form shows `End Date: May 14, End Time: 12:00 AM` for the one that collapsed.

### Cause

`packages/calendar/src/features/calendar/utils/event-form-utils.ts` sets hour and minute but never zeroes seconds or milliseconds:

```ts
export const buildDateTime = (date: Date, time: string, isAllDay: boolean): Dayjs => {
  const [hours, minutes] = time.split(':').map(Number)
  const base = dayjs(date).hour(hours).minute(minutes)   // seconds + ms survive
  return isAllDay ? base.hour(0).minute(0) : base
}
```

The residue comes from the **+ New** button. `openEventForm()` with no arguments seeds `start = currentDate` (`use-calendar-interaction.ts:56`), and `currentDate` originates from `dayjs()` at mount, so it carries the wall-clock seconds and milliseconds of page load. Navigation only does `.add(n, unit)`, which preserves them. The End Date, by contrast, comes from the DatePicker as a clean start-of-day `Date`.

Measured:

```
seeded start -> 2026-05-13 00:00:32.123
picked end   -> 2026-05-14 00:00:00.000
duration     -> 86367 seconds (23:59:27)   end.diff(start, 'day') = 0  -> 1 column

clean start  -> 2026-05-13 00:00:00.000
picked end   -> 2026-05-14 00:00:00.000
duration     -> 86400 seconds (24:00:00)   end.diff(start, 'day') = 1  -> 2 columns
```

`partitionAndSortEvents` (`lib/layout/horizontal.ts:26-31`) classifies an event as multi-unit with `end.diff(start, gridType) > 0`. dayjs truncates, so 23:59:27 becomes `0` and the event takes the single-column path.

Why May 13 in particular: the day of the month happened to match today's, so navigating back to May seeded `currentDate` on the 13th with the page-load time. The other two events had their start date picked from the calendar, which produced a clean `00:00:00.000`. **This means the bug is dependent on the second at which the page loaded**, which is why it looks intermittent.

### Impact beyond the span

Every event created through the form can carry up to 59.999s of invisible duration error. That flows into `DTSTART`/`DTEND` in the iCal export, recurrence `dtstart` and duration math, and any consumer computing durations.

### Fix

Normalize in both builders (`.second(0).millisecond(0)`), and seed the draft in `openEventForm` from a clean minute. Needs a test pinning the duration of a form-built event at exactly 24h, and a span test (see the coverage note at the bottom).

---

### The standard

- **RFC 5545 §3.6.1**: "The DTEND property for a VEVENT calendar component specifies the non-inclusive end of the event." For date-only events it adds that "if such a VEVENT has a DTEND property, it MUST be specified as a DATE value also."
- **Google Calendar API** ([Events resource](https://developers.google.com/workspace/calendar/api/v3/reference/events)): `end` is "The (exclusive) end time of the event."
- **FullCalendar** ([event object](https://fullcalendar.io/docs/event-object)), for reference since consumers migrate from it: "This value is exclusive. For example, an event with the `end` of `2018-09-03` will appear to span through `2018-09-02` but end before the start of `2018-09-03`."

### What is wrong, all measured against current code

**1. Date grids read `end` inclusively.** `lib/layout/horizontal.ts` subtracts a minute only for hour grids:

```ts
const adjustedEnd =
  gridType === 'hour' ? event.end.subtract(1, 'minute') : event.end
```

An event from `2026-05-13T00:00` to `2026-05-14T00:00` covers **two** columns, measured from the layout's own `data-left`/`data-width`. Under the standard it ends as May 14 begins and covers May 13 only.

**2. All-day events are stored ending at `23:59`.** `event-form.tsx` forces `endTime` to `'23:59'` when All day is checked, and `buildEndDateTime` writes `.hour(23).minute(59)`. Day cells do the same: `grid-cell.tsx` emits `endOf('day')` (`23:59:59.999`), which flows into `CellInfo`, the `data-end` attribute the drag-to-create plugin reads, and the form prefill. The conventional value is the next day's midnight.

**3. The iCal export describes the wrong days for all-day events.** Exporting what the form creates today:

| Created in the form | Exported |
|---|---|
| All-day, May 13 only | `DTSTART;VALUE=DATE:20260513` / `DTEND;VALUE=DATE:20260513` |
| All-day, May 13 through May 14 | `DTSTART;VALUE=DATE:20260513` / `DTEND;VALUE=DATE:20260514` |

Because DTEND is non-inclusive, the first describes an event covering no days at all, and the second describes May 13 only. Both are off by one against what the user entered.

**4. `eventOverlapsRange` is inclusive at both ends** (`lib/events/pipeline.ts`), so an event ending exactly at a range start still matches and paints at the boundary.

### Target behavior

| Event | Should cover |
|---|---|
| `2026-05-13T00:00` to `2026-05-14T00:00` | May 13 |
| `2026-05-13T00:00` to `2026-05-14T23:59` | May 13 and May 14 |
| All-day May 13, stored conventionally (`00:00` to May 14 `00:00`) | May 13 |
| All-day May 13, exported | `DTSTART;VALUE=DATE:20260513` / `DTEND;VALUE=DATE:20260514` |

### Notes on Part B

- **This is a breaking behavior change** for consumers passing an inclusive end (for example `end: '2026-05-13T23:59'` for a one-day event). Those keep rendering correctly since `23:59` falls during the end day, but anything relying on the day-grid span or on exported dates shifts. It needs a release note and a migration line, not a silent patch.
- **The form should keep showing an inclusive end date** while storing the exclusive one, so a one-day all-day event still reads "End Date: May 13" rather than May 14. That means initializing the picker from `end.subtract(1, 'day')` for all-day events and converting back on save. This is the split Google uses: exclusive in the API, never a midnight end in the UI.

---

## Shared notes

- **There is no test coverage for multi-day spans.** Flipping the layout line in Part B leaves all 1140 tests passing. Both parts should start by adding the span tests that are missing, so the behavior is pinned rather than assumed.
- **Nothing prevents saving a zero-length event.** `handleEndTimeChange` only validates when start and end fall on the same day, and it compares time strings, so a 12:00 AM end against a 12:00 AM start passes. The result still renders as a full one-day bar.
- Part A is safe to fix on its own and should land first, since it is corrupting stored event data today. Part B is the larger, breaking change and touches the same file.

kcsujeet:
## Linked issue

Closes #248

## What changed?

Both parts of the issue, in two commits so they can be reviewed separately.

### Part A: the form dropped sub-minute precision (`f8fcad2`, bug fix)

`buildDateTime` / `buildEndDateTime` set hour and minute but left anything finer alone, so residue on the source `Date` survived into stored events. The **+ New** path seeds `start` from `currentDate`, which comes from `dayjs()` and carries the seconds of page load; an End Date picked from the calendar arrives clean. Mixing them made a "12:00 AM to next-day 12:00 AM" event **23:59:27** long instead of 24h.

That produced the reported symptom. `layoutHorizontal` classifies multi-day with `end.diff(start, 'day') > 0`, and dayjs truncates, so the event took the single-column path. Three events created the same way rendered differently: May 11 to May 12 and May 19 to May 20 as two-day bars, May 13 to May 14 as one. The difference was whether the start date had been picked from the calendar, which came down to **the second at which the page loaded**.

Both builders now normalize through one `atMinute` helper. `openEventForm` truncates its draft start too, which is not redundant: a consumer's `renderEventForm` receives `selectedEvent` directly and can save it without touching the builders.

### Part B: `end` is now exclusive (`fee3204`, breaking)

| Source | Wording |
|---|---|
| RFC 5545 §3.6.1 | "The DTEND property for a VEVENT calendar component specifies the non-inclusive end of the event" |
| [Google Calendar API](https://developers.google.com/workspace/calendar/api/v3/reference/events) | `end` is "The (exclusive) end time of the event" |
| [FullCalendar](https://fullcalendar.io/docs/event-object) | "an event with the `end` of `2018-09-03` will appear to span through `2018-09-02`" |

Four surfaces now agree with that:

1. **Layout.** `horizontal.ts` steps back a minute at every granularity, not only for hour grids, so an event ending at midnight no longer claims that day's column, exactly as one ending on the hour never claimed that hour's row.
2. **Overlap.** `eventOverlapsRange` stops matching an event whose `end` is the range's first instant, which had put a `00:00`-to-`00:00` event in the following day's list as well. `start` stays inclusive.
3. **All-day storage.** The form writes the midnight after the last covered day rather than `23:59`. The iCal export needed no code change and is now correct: a one-day all-day event used to export `DTEND` equal to `DTSTART`, describing no days at all.
4. **Day cells.** Their range ends at the next midnight, matching the hour and 15-minute branches that already ended on the next slot boundary.

The form keeps showing the **last covered day** and converts on save, so nobody is asked for a midnight end. That is the split Google draws between its exclusive API and its UI. An all-day end that is not on a midnight boundary predates this convention and is read as the inclusive value it is, so existing data keeps displaying correctly.

### Also here: business hours ending at midnight (`e223f22`, bug fix)

Setting the demo's business hours to 12 AM left both time pickers showing "N/A". `getTimeConstraints` builds its max as `(maxEnd - 1):45`, so an end hour of `0` produced `-1:45`, a max earlier than its min, and the option generator returned nothing. The same literal reading made `isBusinessHour` treat every hour as outside business hours, so the grid highlighted nothing either.

`0` and `24` name the same instant, and a picker offering "12 AM" as the closing hour sends the former. `toEndBoundaryMinutes` normalizes it, for end boundaries only, so a `startTime` of `0` still means the start of the day. Business hours of 12 AM to 12 AM now offer all 96 slots, the last being 11:45 PM. An end of 11 PM still stops at 10:45 PM, the last slot that fits before closing.

Three tests used `0` to `0` as shorthand for an empty range and now use `9` to `9`, which is empty under both readings, so their intent is unchanged. A genuinely empty range still yields an "N/A" dropdown; that is a nonsensical configuration rather than the reported one, so it is left alone.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [x] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

## Breaking changes, and what consumers need to do

- A bar that used to reach its `end` day now stops the day before. To keep a two-day bar, end it during the second day (`23:59`) or at the following midnight.
- `CellInfo.end` from `onCellClick`, and the `data-end` attribute plugins read, are now the next midnight rather than `23:59`.
- Inclusive-style ends such as `23:59` still render on the day they name, because that instant falls during the day. Only ends landing exactly on midnight move.

Documented in `apps/website/src/content/docs/help/migration-to-v2.mdx` (three behavior-change entries plus a checklist line), `components/calendar.mdx` (an "`end` is exclusive" section, and its all-day example no longer sets `end` equal to `start`, which under this convention covers no days), and `docs/types-and-interfaces.md`.

## Testing notes

**Multi-day spans had no coverage at all.** Flipping the layout line on its own left all 1140 tests passing, which is exactly why the issue said to write the span tests first. `horizontal.test.ts` now pins four column counts: midnight-to-midnight covers one day, ending during the second day covers two, a conventional three-day all-day event covers three, and a zero-length event still covers one.

**Five existing tests pinned the old inclusive cell range** and were updated deliberately rather than worked around: `isCellDisabled` and `getCellClassName` in `droppable-cell.test.tsx`, the daily cell in `vertical-grid-col.test.tsx` (24 hours, not 23), and the `onCellClick` month-view cases in both calendar test files.

**Part A asserts a duration, not a column count**, on purpose: the column count is what Part B changes, so pinning spans there would only have been rewritten.

The export case builds its event through the form's own helpers, so the two modules stay honest about each other rather than agreeing only in fixtures.

1153 tests pass, `bun run ci` exits 0, and `bunx fallow@2.90.0 audit --changed-since main --gate new-only` is clean across 21 changed files.

## Known consequence, not fixed here

Clicking a month day cell now prefills the form's End Date as the following day at 12:00 AM rather than the same day at 11:59 PM. Both describe the same span under the convention, but the prefill reads oddly. It is a form-affordance question rather than a correctness one, so it is left for a follow-up.


packages/calendar/src/lib/layout/horizontal.ts
```diff
@@ -60,14 +60,23 @@ const computeColumnSpan = (
...
-	const adjustedEnd =
-		gridType === 'hour' ? event.end.subtract(1, 'minute') : event.end
+	// `end` is exclusive at every granularity: an event ending at midnight stops
+	// as that day begins, so it must not claim the day's column, exactly as one
+	// ending on the hour does not claim that hour's row (#248). The step back is a
+	// MILLISECOND, dayjs's smallest unit, so it lands on the last instant the
+	// event occupies and nothing else. A coarser step silently swallows anything
+	// ending inside it: at a minute, an event running 30 seconds into the next day
+	// lost that day entirely.
+	const adjustedEnd = event.end.subtract(1, 'millisecond')
```
kcsujeet, line 78: issue: Partitioning by unit duration drops sub-24h overnight events from the following day.

`partitionAndSortEvents` classifies events with `e.end.diff(e.start, gridType) > 0`, which in Dayjs truncates fractional days to `0`. An overnight event shorter than 24 hours (e.g. 18:00 to 06:00 next morning) gets classified as `singleUnit`, causing `layoutHorizontal` to place it with `endCol: col` on line 212. Even though `computeColumnSpan` here computes the correct 2-column span, the event is clipped to its first day and disappears from the second day in Month View.

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
```diff
@@ -219,9 +219,17 @@ export const EventForm: React.FC<EventFormProps> = ({
...
+	// boundary predates this and is already inclusive, so it is shown untouched.
+	const endsOnMidnight = end.isSame(end.startOf('day'))
+	const lastCoveredDay =
+		initialAllDay && endsOnMidnight ? end.subtract(1, 'day') : end
+
 	// Form state
 	const [startDate, setStartDate] = useState(start.toDate())
-	const [endDate, setEndDate] = useState(end.toDate())
+	const [endDate, setEndDate] = useState(lastCoveredDay.toDate())
 	const [isAllDay, setIsAllDay] = useState(initialAllDay)
```
kcsujeet, line 233: issue: Clicking a month cell and toggling "All Day" creates a two-day event.

Month day cells emit `end: start.add(1, 'day')` with `allDay: undefined`. Because `initialAllDay` is false on draft creation, `lastCoveredDay` leaves `endDate` pointing to the next day instead of stepping back. If the user subsequently checks the "All Day" checkbox and submits, `buildEndDateTime` adds an extra day on save, producing a two-day span instead of one.

packages/plugins/agenda/src/utils/group-events-by-day.ts
```diff
@@ -20,9 +20,10 @@ const appearsOnDay = (
 	dayEnd: Dayjs
 ): boolean => {
 	if (event.allDay) {
-		return (
-			event.start.isSameOrBefore(dayEnd) && event.end.isSameOrAfter(dayStart)
-		)
+		// `end` is exclusive (#248), so an event ending at this day's first instant
+		// covers none of it. Anything looser lists a one-day all-day event under
+		// the following day too, contradicting the grid.
+		return event.start.isSameOrBefore(dayEnd) && event.end.isAfter(dayStart)
```
kcsujeet: issue: Strict `end.isAfter(dayStart)` drops zero-duration all-day events.

For a zero-duration all-day event where `start` and `end` are both `00:00:00`, `event.end.isAfter(dayStart)` evaluates to false on that day's start. This omits the event from the Agenda view entirely, whereas `eventOverlapsRange` and `horizontal.ts` retain and display zero-duration events on their date.

packages/calendar/src/features/calendar/components/event-form/event-form.tsx
```diff
@@ -274,6 +282,21 @@ export const EventForm: React.FC<EventFormProps> = ({
...
+	 * previous day. Without this step back, clicking a month cell (whose range
+	 * ends at the next midnight) and ticking All day saved a two-day event for a
+	 * single day (#248).
+	 */
+	const handleAllDayChange = (checked: boolean) => {
+		setIsAllDay(checked)
+		if (checked && endTime === '00:00') {
+			setEndDate((previous) => dayjs(previous).subtract(1, 'day').toDate())
+		}
+	}
```
kcsujeet, line 307: issue: Checking "All day" when `endDate` is on the same day as `startDate` sets `endDate < startDate`.

If an event starts and ends on the same date with `endTime === '00:00'`, `handleAllDayChange(true)` subtracts 1 day unconditionally, moving `endDate` to the day before `startDate`. On submit, `buildEndDateTime` adds 1 day back to that shifted date, saving a 0-duration event instead of a 1-day event. Guard the subtraction with `dayjs(previous).isAfter(dayjs(startDate))` so it only steps back when `endDate` actually sits on a later day.

packages/plugins/recurrence/src/utils/generate-recurring-events.ts
```diff
@@ -100,12 +100,14 @@ export const generateRecurringEvents = ({
...
-				// Use isSameOrBefore/isSameOrAfter to include boundary cases
+				// Only occurrences that overlap the requested range: the start is
+				// inclusive, the end EXCLUSIVE (#248), so one ending at the range's
+				// first instant covers none of it. This used to read
+				// `end.isSameOrAfter(startDate)`, which contradicted the rule above and
+				// generated that boundary occurrence.
 				const eventSpansRange =
 					recurringEvent.start.isSameOrBefore(endDate) &&
-					recurringEvent.end.isSameOrAfter(startDate)
+					recurringEvent.end.isAfter(startDate)
```
kcsujeet: issue: Strict `recurringEvent.end.isAfter(startDate)` drops zero-duration occurrences at the range start.

For a recurring zero-duration event where `start` and `end` are identical, `end.isAfter(startDate)` evaluates to false when an occurrence lands exactly on `startDate`. This drops that boundary occurrence from the generated instances. Align this with `eventOverlapsRange` so zero-duration occurrences starting on `startDate` are preserved.

PR #252, opened 2026-08-18
feat(dnd): add faithful drag previews

### LINKED ISSUE #251: Feature request: improve timed event drag-and-drop
Currently drag-and-drop only allows to move events to fixed slot durations (1h by default). And this behavior is not too great visually per my taste.

https://github.com/user-attachments/assets/f9fd45d7-fdf5-4b9d-8764-b6f5817a1c3f

I suggest making drag-and-drop behavior closer to Google Calendar, for example.
- Allow events to be moved to sub-hour times, such as 18:45.
- Add a configurable snap interval. For example, a 15-minute interval should
  snap to 18:00, 18:15, 18:30, or 18:45. The default should remain one hour.
- While dragging, optionally show a horizontal guide with the currently
  selected time.
- Make the dragged preview look exactly like the original event. Keep the event
  visible in its original position at half opacity during the drag.

https://github.com/user-attachments/assets/4a9e50ad-a5f9-4b22-a835-92e9d340bf4a

Dreamescaper:
## Linked issue

Closes #251

## What changed?

Improves event drag-and-drop so the floating preview looks and behaves like the event being moved.

- Renders the normal event presentation in the drag overlay, including custom content, colors, truncation, styling, and the source dimensions measured by DnD Kit.
- Keeps the original event visible at half opacity while the floating preview remains full opacity.
- Removes preview enter and drop animations so the event appears immediately at its destination.
- Preserves the grab position for events spanning multiple time slots. Drops resolve from the preview's leading edge and continue to snap according to `slotDuration`.
- Keeps all-day-to-timed drops cursor-anchored because all-day events have no timed leading edge.
- Reads active drag data directly from DnD Kit and removes mirrored overlay state, imperative refs, cursor-centering, and the unused modifiers dependency.

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [x] 🔧 Maintenance

## Verification

- Added regression coverage for default and custom event previews, source opacity, styling updates, guarded drop handling, and multi-slot grab positions.
- Browser-tested middle and bottom grabs on the two-hour `Morning Yoga Class` event in Week view.
- Verified hourly and 15-minute slot durations and all-day conversion fallback.
- `bun run ci` passes with 908 calendar tests.
- Fallow reports no new findings.

## Checklist

- [x] Changes are tested


kcsujeet:
Ran your branch: calendar 932 pass, recurrence 175, drag-to-create 56, agenda 24, type-check clean. Nothing is broken. The issue is scope, see #251.

**Keep:** the overlay work. `DraggableEvent` inside `DragOverlay` at measured size, `@dnd-kit/modifiers` dropped, source at `opacity-50`. That's the real gap, about 40 lines.

**Drop:** `dragSnapInterval` and the `resolveTimedDrag` geometry layer. Snapping stays tied to `slotDuration`; I don't want a second interval that can disagree with the visible grid. Three further reasons the geometry layer shouldn't land:

1. At `slotDuration={15}` in week view its `querySelectorAll` matches 672 cells, each getting a `getBoundingClientRect()` per `onDragMove`.
2. DST bug: `dayOffset` assumes 1440-minute days. In `America/New_York`, a 1h event with its leading edge at 23:58, snap 15, gives `2025-11-03 00:00` on an ordinary day but `2025-11-02 23:00` on the fall-back day. Masked in vertical grids, commits in a resource week hourly lane. See `docs/timezones.md`.
3. Tests mock `@dnd-kit/core` wholesale and stub four hand-built `data-dnd-lane` divs, so nothing pins that the real grid emits those attributes. That matters because `handleDragEnd` skips `fallbackDayDrop` for `time-cell`, making a geometry failure silently inert.

**Split out, own issues:** the recurrence duplicate-override fix (good catch; note the `existingOverride` branch is now unreachable except where it re-creates the duplicate), `getCellRange`'s hardcoded `+15`, and `GridCell` emitting `time-cell`.

Sorry to send back this much. The care is obvious, and those three split-outs are all changes I want.

Dreamescaper:
> Sorry to send back this much. The care is obvious, and those three split-outs are all changes I want.

No worries! I'll drop everything you've requested, and will discuss other functionality in separate issues.

Dreamescaper:
I have reverted per your requests, but I don't like how it behaves without other changes. For events, which span multiple slots (hours), it is quite confusing where event appears when dropped. Dropped slot is resolved from cursor position, which is fine when I'm grabbing the event by the top part, but not by middle or bottom part.


https://github.com/user-attachments/assets/ccce8049-4820-4ec4-b148-b752df699731

Dreamescaper:
> I have reverted per your requests, but I don't like how it behaves without other changes. For events, which span multiple slots (hours), it is quite confusing where event appears when dropped. Dropped slot is resolved from cursor position, which is fine when I'm grabbing the event by the top part, but not by middle or bottom part.

I have added grab position tracking. I can revert it if you wish, but I do think that behavior without this fix is too confusing.

kcsujeet:
Keep it. You're right that the behavior without it is too confusing, so please don't revert `416f6d8`.

Worth naming why it only became a problem now: this PR drops `snapCenterToCursor`, and that modifier is what previously made cursor position ≈ event position. With the overlay rendered at the event's real measured size, the grab offset becomes real, and resolving the drop purely from the cursor puts a tall event somewhere other than where the block appears. That's a consequence of the overlay work I asked for, not something you introduced.

One change to the approach, though. FullCalendar doesn't solve this by measuring rects. It computes a delta between the cell the drag *started* in and the cell under the cursor *now*, and applies that delta to the event's own start — `computeEventMutation` in `packages/preact/src/interaction-plugin/interactions/EventDragging.ts`. The grab offset is preserved by construction, with no geometry involved. The all-day→timed branch of that same function makes the contrast explicit — *"Doesn't matter where on the event the drag began, mutate the event's start-date to date1"* — called out as the exception precisely because the normal timed path is a delta.

I'd rather we follow that. It drops the `slotDuration` parameter, the rect reads and the rounding, and it stays correct when a drag also changes day, resource or row height — cases where dividing by a single cell height won't hold. Two things it needs: `handleDragStart` currently keeps only the event, so the origin cell has to be captured as well (recording the first `over` should be enough); and crossing the all-day boundary needs an absolute set rather than a delta, the same exception FullCalendar carves out.

Heads-up on conflicts: #257 just landed changes to `dnd-utils.ts`, `dnd-utils.test.tsx` and `draggable-event.tsx` — `getUpdatedEvent` gained resource-axis handling in the same block you're editing, and the draggable's data payload gained a new key. Worth rebasing before you rework this so the two don't fight.

Dreamescaper:
Updated - resolved conflicts, updated closer to FullCalendar implementation, dropped `slotDuration` parameter (had to add another parameter `useDestinationTime` though due to `allEventDialog`, where events are not on the grid).

kcsujeet:
Closing this one. I ran the branch against the demo and the full-size overlay does not hold up on the events it has to handle. Screenshots below.

**Week view:** `Research Project` is a timed event running 00:00 to 23:59 across several days, so its source element is the whole 24-hour column. The overlay is that element's measured box, and dnd-kit's `DragOverlay` is `position: fixed`, so nothing clips it, portal or not. The result is a pink column painted over the entire page.

**Month view:** the bar spans every day column it covers and follows the hand, so it overhangs the settings sidebar.

Neither Google Calendar nor FullCalendar float a full-size copy at the cursor. Both snap a mirror of the event into the destination cells, which is a different feature from a `DragOverlay`. Everything else in the diff exists to serve the floating copy: the origin cell, the cell delta and `useDestinationTime` are only needed because a full-size overlay makes the grab offset visible, and a cursor-centered chip has no offset to preserve. The delta also has a DST problem on `day-cell` drags: both midnights sit before the transition while the event's clock time sits after it, so in `America/New_York` a 09:45 event dragged across 2025-11-02 lands at 08:45. FullCalendar avoids that by splitting the delta into whole days plus time of day.

"At measured size" was my ask on 08-19, so this is on me, not you. The work is careful and the split-outs you already made were all improvements.

I'm leaving #251 open and will come back to it later, probably as an in-grid mirror rather than an overlay. Thanks for pushing on this.

Week view: full-height overlay column painted over the page

Month view: multi-day bar overhanging the sidebar

PR #253, opened 2026-08-21
perf: make dayjs comparators answer by instant, and cut the year view to one query

### LINKED ISSUE #245: YearView issues 516 uncached event queries per render, and the per-day hot paths collapse when timezone is set
## Summary

Two independent performance defects, both measured. Neither is a correctness bug
on `main`.

1. `YearView` computes its data in the render body with no memoization, issuing
   **516** `getEventsForDateRange` calls per render.
2. Three per-day hot paths call `startOf`/`endOf` on timezone-aware instances
   inside the event loop. With the `timezone` prop set that costs ~48 µs per
   call, so a single week row with 512 events takes **2.9 seconds** of pure JS.

Filed to give #241 the linked issue the contributing rules require.

## 1. YearView: 516 queries per render

`generateMonthsData` and `generateDaysForMonth` run in the render body:

- 12 month-range queries
- 12 × 42 = 504 mini-calendar day queries

Nothing caches beneath `getEventsForDateRange`, so with
`@ilamy/calendar-recurrence` installed every one of those 516 calls re-expands
every RRULE in the calendar.

Measured on an Apple M4 / Bun 1.3.14, pure JS compute, React reconciliation and
paint excluded:

| Load | Per render, tz cleared | Per render, tz set |
|---|---|---|
| 1 recurring event | 316 ms | 1 299 ms |
| 16 recurring events | 5 444 ms | 17 821 ms |
| 512 plain events | 19.5 ms | 152 ms |

One recurring event is enough to make the year view unusable.

**Memoizing the current shape does not fix it.** All 516 ranges are distinct, so
a range-keyed cache gets zero hits *within* a render — it only helps across
renders, and the first paint is already the problem. The fix has to be
structural: one query for the visible span, bucketed by day.

## 2. The `timezone` prop is a two-orders-of-magnitude cliff

`startOf('day')`, measured against the unpatched library:

| Receiver | `startOf('day')` |
|---|---|
| plain local instance | 352 ns |
| `.utc()` instance | 473 ns |
| **`.tz()` instance** | **47 779 ns** |

This is inherent to `dayjs-timezone`, not to this repo's offset patch. It only
matters because the per-day loops call it inside the event loop, so the cost is
multiplied by events × days:

| Kernel | Load | tz cleared | tz set | Worse by |
|---|---|---|---|---|
| `dayEventsMap`, one 7-day row | 512 events | 15.3 ms | **3 166 ms** | 207x |
| `groupEventsByDay`, 30-day agenda | 512 events | 37.9 ms | **4 392 ms** | 116x |
| `layoutHorizontal`, 42-day month | 512 events | 13.2 ms | 308 ms | 23x |

A 365-day agenda window with 2 048 events had to be dropped from the benchmark
set: with a timezone configured it does not terminate in reasonable time.

### Why this stayed hidden

Each package's `testing-library.ts` clears the default timezone, so a suite runs
timezone-free unless it opts in. Three suites do opt in — `date-utils.test.ts`
(21 `setDefault` calls), `use-calendar-engine.test.ts` (2) and
`recurrence-timezone.test.ts` (2) — but none of them exercise the per-day map,
agenda grouping, or the year-view grid, which is where the cliff lives.

## Reproduction

```tsx
// 1. YearView query count — count calls to getEventsForDateRange through a
//    plugin's transformEvents hook, then render <IlamyCalendar view="year" />
//    with one weekly-recurring event and the recurrence plugin installed.

// 2. Timezone cliff — render any week view with 512 events, once with the
//    timezone prop unset and once set to 'Asia/Tokyo'.
```

The benchmark harness and full methodology (adaptive batching, median of 9
trials, curve fitting) are in the PR discussion.

## Related

Fixed by #241.

## Not part of this issue

Recorded so they are not conflated:

- A consumer inlining `plugins={[...]}` passes a fresh array literal every
  render, which changes the plugin runtime's identity and forces a re-query.
- On `main`, the year-view per-cell event window is skewed by one hour on the
  cells immediately following a DST fall-back, because `getWeekDays` returns an
  instance carrying a stale UTC offset. Day-scale labels are unaffected.

Both are separate, smaller, and out of scope for #241.

kcsujeet:
## Linked issue

Closes #245

## What changed?

**The root cause is one thing, and it is not our code.** Every dayjs comparator routes through `startOf` (`isBefore` → `endOf(units)`; `isSame` → both `startOf` and `endOf`; `isSameOrAfter` = `isSame || isAfter`, so up to three), and dayjs-timezone overrides `startOf`:

```js
proto.startOf = function (units, startOf) {
  if (!this.$x || !this.$x.$timezone) return oldStartOf.call(this, units, startOf)
  var withoutTz = d(this.format('YYYY-MM-DD HH:mm:ss:SSS'), { locale: this.$L })
  return oldStartOf.call(withoutTz, units, startOf).tz(this.$x.$timezone, true)
}
```

Format → reparse → re-anchor, on every call, including the no-op `startOf(undefined)`. Two consequences:

- **Cost.** 45–135 µs per comparison with a `timezone` set, multiplied by events × days on every per-day filter.
- **Correctness.** The re-anchor *moves* an instance carrying a stale UTC offset, which `add()` produces across a DST transition, so two instants an hour apart compared equal.

Both are open upstream and neither identifies the cause: [dayjs#1399](https://github.com/iamkun/dayjs/issues/1399) (correctness, 2021) and [dayjs#2344](https://github.com/iamkun/dayjs/issues/2344) (performance, 2023).

### 1. `compareByInstant`

`isBefore`, `isAfter`, `isSame`, `isSameOrBefore`, `isSameOrAfter` now answer by epoch millisecond when no unit is given **and** the operand is already a Dayjs — which is what dayjs documents unitless comparison to be ("[default milliseconds](https://day.js.org/docs/en/query/is-before)"). A unit asks a different question, and any other operand needs dayjs's own coercion; both fall through untouched.

| | before | after |
|---|---|---|
| `a.isBefore(b)` | 44 957 ns | **10 ns** |
| `a.isSameOrAfter(b)` | 134 739 ns | **13 ns** |
| `a.isSame(b)` | 64 794 ns | **13 ns** |
| `a.isSame(b, 'day')` *(falls back)* | — | 101 684 ns |
| `a.isBefore('2025-01-01')` *(falls back)* | — | 49 282 ns |

```
stale        : 2025-03-10 00:00 -05:00   (05:00Z)
true midnight: 2025-03-10 00:00 -04:00   (04:00Z)

                              before    after
stale.isSame(trueMidnight)  →  true   →  false
stale.isAfter(trueMidnight) →  false  →  true
```

### 2. `dayKey` / `isSameDay`

The one comparison worth keeping our own. `isSame(x, 'day')` carries a unit, so it still routes through dayjs-timezone's `startOf` at ~101 µs, and every grid cell asks it. Rendered keys answer for ~1.5 µs. `isToday` uses them, and `AgendaDayGroup`'s inline copy of `isToday` now shares them.

### 3. YearView: 516 queries per render → 1

`generateMonthsData` and `generateDaysForMonth` ran in the render body with no memoization: 12 month ranges + 12 × 42 mini-calendar days. Nothing caches beneath `getEventsForDateRange`, so with the recurrence plugin every one re-expanded every RRULE. Observed through a plugin's `transformEvents` the count was **1034** (517 × two render passes); it is now **4**, pinned with `toBe`.

Memoizing the old shape would not have helped: all 516 ranges are distinct, so a range-keyed cache gets no hits within a render.

### 4. dayjs 1.11.21 → 1.11.23, `isBetween` dropped

Two timezone-plugin fixes ([#3174](https://github.com/iamkun/dayjs/pull/3174), [#3180](https://github.com/iamkun/dayjs/issues/3180)). `isBetween` had zero references anywhere, tests included.

## Measured, end to end

| Year view render | tz | before | after |
|---|---|---|---|
| 1 recurring event | cleared | 126.4 ms | 62.3 ms |
| 16 recurring events | cleared | 819.4 ms | 54.3 ms |
| 512 plain events | cleared | 617.2 ms | 35.2 ms |
| 1 recurring event | New York | 655.3 ms | 163.3 ms |
| 16 recurring events | New York | 5 655 ms | 502 ms |
| 512 plain events | New York | **146 632 ms** | **126 ms** |

Agenda grouping, 30-day window, 512 events: 2 490 ms → 6.8 ms (New York), **with `group-events-by-day.ts` unchanged**.

## What is deliberately NOT in this diff

This change was twice as large before it was right. The first version added numeric comparators to `@ilamy/utils/helpers` and rewrote every hot call site: the agenda's `appearsOnDay` and day loop, `layout/vertical.ts`, `layout/horizontal.ts`, the recurrence override lookup, `overlapsRange` itself. Once `compareByInstant` existed, all of it was redundant — `overlapsRange` measures **124 ns using plain dayjs comparators against 99 ns hand-rolled**, with zero disagreements across 49 boundary cases.

Fixing the layer that was actually broken removed the need to touch those files, and the riskiest edit went with them: removing `cursor.startOf('day')` from the agenda loop as "dead work" had introduced a DST bug, since `add(1, 'day')` holds the wall clock but carries the previous offset across a transition. That file is now untouched. Its five new timezone/DST tests are kept, because no suite covered agenda grouping under a timezone and that gap was real.

## `fixTimezoneOffset` cannot be removed

That was the hope behind the upgrade, since [#3169](https://github.com/iamkun/dayjs/issues/3169)'s root cause ("both `new Date(target)` and `getTimezoneOffset()` depend on the host") is the same family. Tested against **raw upstream 1.11.23 with no patch** — all 365 days of 2025, six zones, `startOf('day')` and `endOf('day')`, offsets checked against `Intl`:

| host TZ | checks | mismatches |
|---|---|---|
| UTC | 4 380 | 4 |
| Europe/Paris | 4 380 | 4 |
| Europe/London | 4 380 | 4 |
| America/New_York | 4 380 | 12 |
| Australia/Sydney | 4 380 | 0 |

Always on a DST transition day. With the patch: **0 on every host tried.**

## Tests written before the flip, deliberately

An earlier prototype of `compareByInstant` passed all 1 195 tests while silently changing `isSame` from true to false on drifted instances — exactly the shape `.agents/rules/change-impact.md` warns about. Two of the six new `dayjs.test.ts` cases fail against the old implementation.

## Type of Change

- [x] 🐛 Bug fix
- [x] 💥 Breaking change (see below)
- [x] 🔧 Maintenance

## Breaking

Two changelog lines needed:

- Unitless comparison now answers by instant, so a Dayjs carrying a stale UTC offset compares differently. The new answer matches dayjs's documented contract.
- `.isBetween()` is no longer available on the `dayjs` exported from the Plugin SDK.

## Relationship to #241

Same diagnosis, different implementation, and credit for the diagnosis belongs to @mattanderson-io. All figures here were re-measured locally.

Deliberately not taken from that branch: the `@ilamy/utils/events` module, the `WeakMap` bounds cache, `buildDayIndex`/`daySpanOf`, and the memoization of the DST offset check. The cache exists there to avoid recomputing `startMs`, but `valueOf()` costs 4 ns. Both of its review regressions (gapped day lists, year-view DST grid shift) came from the index it introduced, so avoiding the index avoids that class.

Net: **18 files** against that branch's 20, and the substantive fix is one dayjs plugin rather than a 352-line module.

## Also out of scope

- **The year-view cells keep their un-normalized `dayDate`.** After a fall-back its label, click target and event window all derive from the same skewed instant and agree with each other. #241 noted the one-hour window skew as a separate defect; normalizing only some derivations is what shifted its grid by a whole day.
- **The remaining ~72 ms** of 504 tz-aware `startOf`/`endOf` pairs in the year view. `main` paid that too.

## Checklist

- [x] CI passes locally (`bun run ci`) — 1 193 tests, 0 failures
- [x] Fallow gate clean on all 18 changed files
- [x] Changes are tested
- [x] `docs/logs/2026-08-21.md` added


kcsujeet (commented):
Reviewed all 23 changed files. The instant comparison plugin, dayKey/isSameDay helpers, and YearView single-query memoization are clean and correctly preserve timezone boundaries. CI and fallow pass.

Minor docs follow-up: packages/utils/README.md still lists isBetween in its plugin list and can be updated.

PR #254, opened 2026-08-21
fix(dayjs): compare unitless by instant, and upgrade to 1.11.23

kcsujeet:
## Linked issue

Follow-up to #245, split out of #253 as agreed there.

> **Stacked on #253.** Base branch is `perf/numeric-hot-path-comparisons`, not `main`. Review #253 first; this diff shows only the delta.

### 1. dayjs comparators answer by instant

dayjs documents unitless comparison as ["default milliseconds"](https://day.js.org/docs/en/query/is-before). It does not deliver that for timezone-aware instances. Every comparator routes through `startOf` (`isBefore` → `endOf(units)`; `isSame` → both `startOf` and `endOf`; `isSameOrAfter` = `isSame || isAfter`, so up to three), and dayjs-timezone overrides `startOf` with a format → reparse → `.tz(zone, true)` round-trip that runs even for the no-op `startOf(undefined)`.

That re-anchor **moves** an instance carrying a stale UTC offset, which `add()` produces across a DST transition:

```
stale        : 2025-03-10 00:00 -05:00   (05:00Z)
true midnight: 2025-03-10 00:00 -04:00   (04:00Z)

                              before    after
stale.isSame(trueMidnight)  →  true   →  false
stale.isAfter(trueMidnight) →  false  →  true
```

Two instants an hour apart comparing equal is a correctness bug, not a preference. The new `compareByInstant` plugin fixes it and removes the cost at the same time.

The fast path applies **only** when no unit is given and the operand is already a Dayjs. A unit asks a different question, and any other operand needs dayjs's own coercion; both fall through untouched.

| | before | after |
|---|---|---|
| `a.isBefore(b)` | 44 957 ns | **10 ns** |
| `a.isSameOrAfter(b)` | 134 739 ns | **13 ns** |
| `a.isSame(b)` | 64 794 ns | **13 ns** |
| `a.isSame(b, 'day')` *(falls back)* | — | 101 684 ns |
| `a.isBefore('2025-01-01')` *(falls back)* | — | 49 282 ns |

Both halves are reported upstream and still open: [dayjs#1399](https://github.com/iamkun/dayjs/issues/1399) (correctness, 2021) and [dayjs#2344](https://github.com/iamkun/dayjs/issues/2344) (performance, 2023). Neither identifies the cause; a local fix is the only realistic option.

### 2. dayjs 1.11.21 → 1.11.23

Two timezone-plugin fixes: 1.11.22 "compute instance `.tz()` offset without host DST" ([#3174](https://github.com/iamkun/dayjs/pull/3174), closing [#3169](https://github.com/iamkun/dayjs/issues/3169)) and 1.11.23 "prevent RangeError for invalid Day.js values" ([#3180](https://github.com/iamkun/dayjs/issues/3180)). Bumped in all five packages that declare it.

## `fixTimezoneOffset` cannot be removed

That was the hope behind the upgrade, since #3169's root cause ("both `new Date(target)` and `getTimezoneOffset()` depend on the host") is the same family as the bug our patch works around. Tested rather than assumed.

Against **raw upstream 1.11.23 with no patch**: walking all 365 days of 2025 in six zones (`America/New_York`, `Europe/London`, `Europe/Berlin`, `America/Guadeloupe`, `Australia/Lord_Howe`, `Asia/Kathmandu`), calling `startOf('day')` and `endOf('day')`, and checking the reported offset against `Intl.DateTimeFormat`:

| host TZ | checks | mismatches |
|---|---|---|
| UTC | 4 380 | 4 |
| Europe/Paris | 4 380 | 4 |
| Europe/London | 4 380 | 4 |
| America/New_York | 4 380 | 12 |
| Australia/Sydney | 4 380 | 0 |

Always on a DST transition day, e.g. `Europe/London startOf -> 2025-03-30 00:00 +01:00 got=60 want=0`. With our patch: **0 mismatches on every host tried**. The patch is load-bearing. The upgrade is still worth taking on its own merits.

## Tests written before the flip, deliberately

The earlier prototype of this change passed all 1 195 tests while silently flipping `isSame` from true to false on drifted instances. That is exactly the shape `.agents/rules/change-impact.md` warns about: behaviour nobody had pinned. So `dayjs.test.ts` gained six cases first, two of which fail against the old implementation:

- two instants an hour apart are different
- a stale-offset instance orders by its real instant
- ordinary instances agree with `valueOf`
- unit-ed comparison still delegates to dayjs
- string, `Date` and number operands still work

## Type of Change

- [x] 🐛 Bug fix
- [x] 💥 Breaking change (see below)
- [x] 🔧 Maintenance

## Breaking

`isSame`/`isBefore`/`isAfter`/`isSameOrBefore`/`isSameOrAfter` change answers for Dayjs operands carrying a stale UTC offset. The new answer matches dayjs's documented contract, but consumers relying on the old re-anchoring behaviour through the publicly exported `dayjs` would see a difference. Worth a changelog line alongside the `isBetween` removal from #253.

## Open question for review

With dayjs's own comparators now at ~12 ns, the **instant** helpers added to `@ilamy/utils/helpers` in #253 (`isBefore`, `isAfter`, `isSame`, `isSameOrBefore`, `isSameOrAfter`, `isBetweenInclusive`) are no longer a performance necessity, only a stylistic one. Happy to drop them and revert those call sites to plain dayjs methods if you prefer a smaller API surface.

`dayKey` and `isSameDay` should stay either way: `isSame(x, 'day')` carries a unit, so it still falls back to dayjs at ~101 µs.

## Checklist

- [x] CI passes locally (`bun run ci`) — 1 200 tests, 0 failures
- [x] Fallow gate clean on all 22 changed files
- [x] Changes are tested
- [x] `docs/logs/2026-08-21.md` updated


kcsujeet:
Folded into #253.

Splitting these was the wrong call and the measurements are why: once the dayjs comparator patch from this PR exists, most of #253 stops being necessary. `overlapsRange` costs 124 ns using plain dayjs comparators against 99 ns hand-rolled, with zero disagreements across 49 boundary cases, so the agenda grouping, both layout modules, the recurrence override lookup and the six helper comparators all needed no change at all.

Rather than hand over an add-then-remove history, #253 is now a single commit containing only what survives: `compareByInstant`, `dayKey`/`isSameDay`, the YearView one-query fix, the 1.11.23 upgrade and the `isBetween` removal. 18 files instead of 22, and the two riskiest edits are gone with it.

PR #256, opened 2026-08-22
fix(calendar): forward slotDurationMinutes to DroppableCell

### LINKED ISSUE #255: slotDuration is ignored when computing a cell's end time (CellInfo.end always 15 min short at slotDuration=30)
## Bug

With `slotDuration={30}`, every cell's end time is always 15 minutes short of where it should be, not proportional to how long you drag, just always exactly 15 minutes off. So dragging or clicking a 7:00–8:30 range creates a 7:00–8:15 event instead. A 2-hour drag comes up 15 minutes short too.

## Root cause

`slotDuration` gets passed down correctly almost all the way except the last step. `GridCell` (`grid-cell.tsx`) receives the right `slotDurationMinutes`, but never passes it on to `DroppableCell` (`droppable-cell.tsx`). `DroppableCell` computes each cell's own end time, and hardcodes it to `+15 minutes` no matter what `slotDuration` actually is:

```ts
// packages/calendar/src/components/droppable-cell.tsx
if (hour !== undefined && minute !== undefined) {
	return { start, end: start.minute(minute + 15) } // always +15, ignores slotDuration
}
```

This is the exact value used for `onCellClick` and the drag-to-create plugin, so it's what causes the wrong end time. Looks like it got missed in #140, the PR that added `slotDuration` in the first place — that PR never touched either of these two files.

## Repro

```tsx
<IlamyCalendar
  orientation="vertical"
  resources={resources}
  slotDuration={30}
  onCellClick={(info) => console.log(info.end.toISOString())}
/>
```

Click the 7:30 cell — logs 7:45 instead of 8:00.

## Fix

Pass `slotDurationMinutes` from `GridCell` down to `DroppableCell`, and use it instead of the hardcoded `15`. I have this ready, verified against the existing test suite plus a new test case — opening a PR linked to this issue.

johanjq:
Closes #255

## What's wrong

`slotDuration` is threaded correctly from `IlamyCalendar` down through `VerticalGrid` → `VerticalGridCol` → `GridCell`, which uses it for the business-hours containment check — but `GridCell` never forwards it to the `DroppableCell` it renders. `DroppableCell` computes each minute-cell's own end time and hardcodes it to `+15 minutes`, no matter what `slotDuration` is actually configured. That's the value `onCellClick` and the drag-to-create plugin both read, so at `slotDuration={30}` every click/drag reports an end time 15 minutes short of where it should be — a flat offset regardless of drag length.

This looks like it was missed in #140 (the PR that added `slotDuration`) — it never touched `droppable-cell.tsx` or `grid-cell.tsx`.

## What changed

- `droppable-cell.tsx`: added a `slotDurationMinutes` prop, used by `getCellRange()` instead of the hardcoded `15` (default stays `15`, so existing behavior is unchanged unless the prop is passed).
- `grid-cell.tsx`: forwards `slotDurationMinutes` to `<DroppableCell>` — the one missing link in the chain.
- `droppable-cell.test.tsx`: added a case covering `slotDurationMinutes={30}`.

## Testing

```
bun test src/components/droppable-cell.test.tsx
 22 pass
 0 fail
```

Also ran the full `packages/calendar` suite before and after this change to confirm no regressions (the 6 failures present are pre-existing on `main`, unrelated to this change — a workspace build-order issue with `@ilamy/calendar` not being built yet).


packages/calendar/src/components/droppable-cell.test.tsx
```diff
@@ -194,6 +196,18 @@ describe('DroppableCell self-describing attributes (drag-create)', () => {
 		)
 	})
 
+	test('honors slotDurationMinutes on a minute cell', () => {
```
kcsujeet, line 199: issue: this test passes with or without the fix

`renderCell` renders `DroppableCell` directly and hands it `slotDurationMinutes` itself. The defect was `GridCell` never forwarding it, so deleting `slotDurationMinutes={slotDurationMinutes}` from `grid-cell.tsx` again leaves this green. It pins the new prop, not the bug.

The regression needs the real entry point:

```tsx
render(
  <IlamyCalendar events={[]} initialView="day" slotDuration={30}
    initialDate={dayjs('2025-07-15T00:00:00.000Z')} />
)
// 07:00 cell data-end: 07:15 on main, 07:30 with this fix
```

packages/calendar/src/components/droppable-cell.tsx
```diff
@@ -22,18 +28,20 @@ interface DroppableCellProps {
...
+ * The time span a cell represents. Granularity follows the view: a
+ * `slotDurationMinutes`-wide slot (day/resource grids), a one-hour slot
+ * (week), or the whole day (month).
  */
 function getCellRange(
 	date: Dayjs,
 	hour?: number,
-	minute?: number
+	minute?: number,
+	slotDurationMinutes = 15
```
kcsujeet: issue: the `= 15` default is dead, and it re-arms the bug

`GridCell` is the only production caller of `DroppableCell`, so after this change nothing reaches this default. What it does do is let a future caller omit the prop and silently get 15-minute cells again, which is exactly how this slipped through #140.

Making `slotDurationMinutes` required lets TypeScript enforce the wiring instead. There is no value that is safe to default to here.

kcsujeet (commented):
Confirmed the diagnosis independently before reviewing. Rendering `IlamyCalendar` on `main` and reading `data-end` off the 7am cells:

| `slotDuration` | 7am cells |
|---|---|
| 60 | `07:00->08:00` (60m) |
| 30 | `07:00->07:15`, `07:30->07:45` (15m) |
| 15 | `07:00->07:15` ... (15m) |

So only `slotDuration={30}` is affected: at 15 the hardcoded value is right by accident, and 60 takes the hour branch. Worth narrowing the issue description to that, since "always 15 minutes short" reads as if it hits every configuration.

Two inline comments. One more thing that has no line to hang off: AGENTS.md requires a `docs/logs/` entry for any change under `src/`, and this PR has none. Worth a short entry noting that `CellInfo.end` and `data-end` change at `slotDuration={30}`, since both are public surface (`onCellClick`, `isCellDisabled`, `getCellClassName`, and the drag-to-create plugin read them).

Also: the "6 pre-existing failures on `main`" you saw are a build-order artifact rather than a real baseline. The plugin packages resolve `@ilamy/calendar` through its built `dist`, so a bare `bun test` on a clean tree fails on module resolution. `bun run ci` builds first. I hit the same thing this week and briefly took it for a regression.

No conflict with #252, by the way. That PR was rescoped and no longer touches either file.

packages/calendar/src/components/droppable-cell.test.tsx
johanjq: sounds good 👌🏼

packages/calendar/src/components/droppable-cell.tsx
johanjq: yeah I have removed this now and it's required! thanks.

kcsujeet (approved):
Verified the fix by reverting it and watching the new test fail with the original bug. CI green.

kcsujeet:
Merged, thanks for this one and for turning the test around so quickly. It'll go out in the next release, and I'll update here when that's published.

PR #260, opened 2026-09-02
fix: keep cells interactive in weeks that span two months

### LINKED ISSUE #259: Week view disables cells that fall in the neighbouring month
## What happens

Any week that spans two months comes up half interactive. In the week of
Monday 31 March 2025 (which runs to Sunday 6 April), if you arrived from
March you can click, drag into, and drop onto Monday 31 — but 1–6 April are
greyed out and inert. Navigate to the same week from April and it reverses:
April works, 31 March is dead.

The affected cells:

- refuse cell clicks, so you cannot create an event on them
- reject drops, so you cannot drag an existing event onto them
- render with the disabled styling

The week's all-day row behaves the same way, as do the resource week grids.

## Expected

Every day shown in week view is a day the view meant to show. None of them is
padding, so all of them should be interactive — subject only to
`businessHours` and `isCellDisabled`, which are separate opt-in rules.

## Reproduce

```tsx
<IlamyCalendar initialView="week" initialDate={dayjs('2025-03-31')} />
```

Try to click a cell on Tuesday 1 April. Nothing happens. Switch
`initialDate` to `2025-04-01` and Monday 31 March goes dead instead.

Roughly five or six weeks a year straddle a boundary, so on any given month
this is one or two weeks out of five.

## Cause

`GridCell` computes `isCurrentMonth = day.month() === currentDate.month()`
and passes `disabled={!isBusiness || !isCurrentMonth}`
(`packages/calendar/src/components/grid-cell.tsx`).

That rule belongs to the month grid, whose first and last rows are padded with
days from the neighbouring months so every week is seven cells wide. `GridCell`
is shared by every grid, so the rule is applied everywhere — including the grids
that have no padding.

Day view is unaffected: it derives every cell from `currentDate`, so the months
always match.

## No workaround from the outside

`isCellDisabled` is combined with this check as
`disabled || Boolean(isCellDisabled?.(cellInfo))`, so a consumer predicate can
only *add* disabled cells. There is no prop that restores one.

I'll open a PR soon with the patch I'm currently using to fix it.

## Version

`@ilamy/calendar` 2.0.3, and still present on `main` at b5b4514.

habovh:
## Linked issue

Closes #259 

## The rule that leaked

`GridCell` disabled any cell whose month differed from `currentDate`'s:

```ts
const isCurrentMonth = day.month() === currentDate.month()
// ...
disabled={!isBusiness || !isCurrentMonth}
```

That describes the **month** grid. Its first and last rows are padded with days
from the neighbouring months so every week is seven cells wide, and those cells
are context rather than part of the month being edited.

But `GridCell` is shared by every grid, so the rule ran everywhere — including
grids that have no padding at all.

## What it did to week view

Every week that spans a month boundary came up half dead. In the week of Monday
31 March 2025, arriving from March you could click and drop on Monday 31 but
not on 1–6 April; arriving from April reversed which half worked. The cells
refused clicks, rejected drops, and rendered greyed out.

The week's all-day row went the same way, and so did the resource week grids —
they build their cells from the same `getWeekDays` result. Day view was never
affected: it derives every cell from `currentDate`, so the months always
matched.

Consumers could not work around it. `isCellDisabled` is combined as
`disabled || Boolean(isCellDisabled?.(cellInfo))`, so a predicate can only add
disabled cells, never restore one.

## What changed?

One line of behaviour: scope the rule to the grid that actually pads itself.

```ts
const isOutsideDisplayedMonth =
  view === 'month' && day.month() !== currentDate.month()
```

`view` comes from the same context `GridCell` already reads `currentDate` from,
and in production the rendered view is resolved *from* it — `IlamyCalendar`
does `getViews().find((v) => v.name === view)` — so the two cannot disagree.

Month view keeps its behaviour exactly. Business hours are untouched and still
disable cells on their own terms; the two reasons are now independent, which
one of the tests pins directly.

## Tests

Five, all failing before the change and passing after (verified by reverting
the one line and watching them go red):

- week view, both directions across the boundary — because `currentDate` can
  sit on either side depending on how you navigated in, and a fix that only
  handled one direction would look right half the time
- week view's all-day row across the boundary
- week view, a next-month cell outside business hours stays disabled — proving
  the fix did not weaken the business-hours path
- month view, padded cells from both neighbouring months stay disabled, while a
  day the month owns stays interactive — the regression guard for the behaviour
  that is deliberately kept

I also added `initialView="week"` to the week test harness. It mounted the week
grid while the context still reported `'month'`, so anything reading the active
view saw the wrong one — the new tests could not even engage until the harness
matched production. Worth a look: it may be masking other view-dependent
behaviour in that file.

## An alternative you may prefer

Reading `view` inside a shared leaf component is the smallest change, but it
does mean `GridCell` knows a view name. The alternative is to let the producer
say so: add an optional `outsidePeriod` to `HorizontalCellSpec`, set it in
`month.tsx`, and have `GridCell` take it as a prop. That removes the implicit
invariant entirely and would extend to any plugin view that pads its grid, at
the cost of a new (additive) field on a public type.

I went with the smaller one for a bug fix, but I'm happy to switch if you'd
rather the knowledge lived with the view that creates the padding.

## AI disclaimer

I've used Claude Opus 5 to help on this PR.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

### DIFF
--- a/packages/calendar/src/components/grid-cell.tsx
+++ b/packages/calendar/src/components/grid-cell.tsx
@@ -61,6 +61,7 @@ const NoMemoGridCell: React.FC<GridProps> = ({
 		eventSpacing,
 		eventHeight,
 		onMoreEventsClick,
+		view,
 	} = useSmartCalendarContext()
 	const effectiveBusinessHours = useEffectiveBusinessHours(resourceId)
 
@@ -112,7 +113,14 @@ const NoMemoGridCell: React.FC<GridProps> = ({
 		allEventsDialogRef.current?.open()
 	}
 
-	const isCurrentMonth = day.month() === currentDate.month()
+	// Only the month grid pads itself: its first and last rows carry days from
+	// the neighbouring months so every week is seven cells wide, and those are
+	// context rather than part of the month being edited. Every other grid
+	// shows exactly the days it means to, and those days are free to cross a
+	// month boundary — the week of 31 March runs into April, and April is not
+	// padding there.
+	const isOutsideDisplayedMonth =
+		view === 'month' && day.month() !== currentDate.month()
 
 	const hiddenEventsCount = todayEvents.length - dayMaxEvents
 	const hasHiddenEvents = hiddenEventsCount > 0
@@ -150,7 +158,7 @@ const NoMemoGridCell: React.FC<GridProps> = ({
 				)}
 				data-testid={dataTestId || testId}
 				date={day}
-				disabled={!isBusiness || !isCurrentMonth}
+				disabled={!isBusiness || isOutsideDisplayedMonth}
 				hour={hour}
 				id={droppableId}
 				minute={minute}
--- a/packages/calendar/src/features/calendar/components/views/month.test.tsx
+++ b/packages/calendar/src/features/calendar/components/views/month.test.tsx
@@ -236,6 +236,25 @@ describe('MonthView', () => {
 			expect(within(dialog).getByText(title)).toBeInTheDocument()
 		})
 	})
+
+	// The month grid pads its first and last rows with days from the
+	// neighbouring months so every week is seven cells wide. Those cells are
+	// context, not part of the month being edited, and stay disabled. Pinned
+	// because the same check used to run in week and day view, where the days
+	// are not padding at all.
+	test('disables the days padded in from the neighbouring months', () => {
+		cleanup()
+		// March 2025 starts on a Saturday, so the first row is padded with
+		// 23-28 February and the last with 1-5 April.
+		renderMonthView({ initialDate: dayjs('2025-03-15T00:00:00.000Z') })
+
+		const cellOn = (date: string) => screen.getByTestId(`day-cell-${date}`)
+
+		expect(cellOn('2025-02-26').dataset.disabled).toBe('true')
+		expect(cellOn('2025-04-02').dataset.disabled).toBe('true')
+		// A day the month actually owns stays interactive.
+		expect(cellOn('2025-03-12').dataset.disabled).toBe('false')
+	})
 })
 
 const resourceList: Resource[] = [
--- a/packages/calendar/src/features/calendar/components/views/week.test.tsx
+++ b/packages/calendar/src/features/calendar/components/views/week.test.tsx
@@ -54,6 +54,11 @@ const renderWeekView = (props = {}) => {
 			events={mockEvents}
 			firstDayOfWeek={firstDayOfWeek}
 			initialDate={initialDate}
+			// Mirrors production, where the rendered view is resolved *from*
+			// `view` on the context (see IlamyCalendar). Without it the harness
+			// mounts the week grid while the context still says 'month', and
+			// anything that reads the active view sees the wrong one.
+			initialView="week"
 			locale={locale}
 			{...props}
 		>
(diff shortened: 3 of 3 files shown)


habovh:
I've pushed a second commit with the alternative suggested earlier with a new `outsidePeriodBehavior` prop controlling what happens when interacting with out-of-period days, with 3 possible values and only affects month view:
- `disabled` mimics the current behavior where nothing happens
- `interactive` where the out-of-period days behave the same way as normal days
- `navigate` where clicking on an out-of-period day navigates to the corresponding month.

Of course this is optional, and expands a bit on the initial scope of the PR so this could be split into two PRs if you prefer, but since they are loosely related and in my opinion improve the overall experience, I took the opportunity to bundle the initial fix with this new prop together.

habovh:
Fallow was right, and the fix is pushed.

`NoMemoGridCell` was at CRAP 49.5 against a threshold of 30. CRAP squares
cyclomatic complexity, and this branch had walked that from 7 to 13 by adding
one derived boolean at a time across three commits — disabled-by-period,
navigates-on-click, and the class that belongs to neither.

They were always one decision, so they now read as one.
`outsidePeriodPresentation` answers what being outside the period does to a
cell, in the order the rules actually apply: inside the period, nothing;
disabled, no class of its own because `disabledCell` already covers it;
otherwise interactive, marked, and navigating only when asked.

`NoMemoGridCell` is back to cyclomatic 8 against main's 7, which clears the
gate. No behaviour change — the outside-period tests from the previous commit
pass untouched, and the full suite is green.

I reproduced this locally with `fallow-darwin-arm64` v3.20.0, the same version
the workflow pins, rather than guessing at the metric:

```
before   NoMemoGridCell  cyclo 13  cog 24  CRAP 49.5  introduced: true   → fail
after    NoMemoGridCell  cyclo  8  cog 19             introduced: false  → pass
```

## On the second finding

`DroppableCell` is reported alongside it at cognitive 21. That one is
`introduced: false` — it was 20 on `main`, already over the threshold of 15
before this branch existed, and the gate excludes it. This branch adds one
point by passing it a prop.

I have left it alone. Bringing it under 15 means restructuring a function this
branch only passes a prop to, which is a different change from this one and
would be better on its own. Happy to do that separately if you want it.

kcsujeet (commented):
I think the whole PR is more complicated than it needs to be. The first commit was simple and worked well enough to address the bug. 

```js
const isOutsideDisplayedMonth =
		view === 'month' && day.month() !== currentDate.month()
```

habovh:
@kcsujeet alright, there definitely has been some scope creep here. I'll update the branch with just the week view fix. Would you be open to the control suggestions for month view in a separate feature PR?

kcsujeet (approved):
Thanks for cutting this back, this is the shape I wanted.

On the month-view controls: not taking them. Neither reference has the concept. FullCalendar has `showNonCurrentDates` (a render toggle) and an `isOther` styling flag, no behaviour mode and nothing that makes a click navigate instead of firing the cell handler. Google Calendar publishes nothing on it.

PR #262, opened 2026-09-02
fix: give DndContext a stable id so server-rendered calendars hydrate

### LINKED ISSUE #123: The "now" line renders in day-resolution vertical views, likely not needed

- **Resource Month Vertical**

entirely?

clearly.

### LINKED ISSUE #261: Server-rendered calendars fail to hydrate after the first render
## What happens

Any app that server-renders the calendar (Next.js, Remix, TanStack Start)
throws a hydration error on every page load **except the first one the server
process handles**:

```
A tree hydrated but some attributes of the server rendered HTML didn't match
the client properties.

+ aria-describedby="DndDescribedBy-0"     (client)
- aria-describedby="DndDescribedBy-1"     (server)
```

## Cause

`@dnd-kit/utilities` numbers its ids from a module-level counter that is never
reset:

```js
let ids = {}
function useUniqueId(prefix, value) {
  if (value) return value
  const id = ids[prefix] == null ? 0 : ids[prefix] + 1
  ids[prefix] = id
  return prefix + '-' + id
}
```

In a browser that is fine — the module is fresh on every page load. On a server
it is not: the module lives for the life of the Node process, so the counter
carries from one request to the next. Fetching a calendar page four times in a
row returns `DndDescribedBy-6`, `-7`, `-8`, `-9`, while the client always starts
over at 0.

The attribute is on every draggable event, so a calendar with no events does
not reproduce it.

## Reproduce

Two `renderToString` calls in one process are exactly two requests:

```tsx
const markup = () => renderToString(<IlamyCalendar events={[oneEvent]} />)
markup() // aria-describedby="DndDescribedBy-0"
markup() // aria-describedby="DndDescribedBy-1"
```

## Impact

The first load after a server restart is clean, which is what makes this
confusing to report — in development it looks intermittent, and in production
it is permanent for every visitor after the first.

Consumers cannot work around it: the `DndContext` is internal, so there is no
`id` to pass. The only workaround is to stop server-rendering the calendar.

## Version

`@ilamy/calendar` 2.0.3, and still present on `main` at b5b4514.

habovh:
## Linked issue

Closes #261 

## The bug

`@dnd-kit/utilities` numbers the ids behind `aria-describedby` from a
module-level counter that is never reset:

```js
let ids = {}
const id = ids[prefix] == null ? 0 : ids[prefix] + 1
```

In a browser that is fine — the module is fresh on every page load. On a server
it is not. The module lives for the life of the process, so the counter carries
from one request to the next: the fourth render of a page emits
`DndDescribedBy-3` while the client, starting over at 0, emits
`DndDescribedBy-0`. Every server-rendered calendar after the process's first one
fails to hydrate.

The first load after a server restart is clean, which is why this reads as
intermittent in development and is permanent in production.

## The fix

`useUniqueId(prefix, value)` returns `value` untouched when one is given, and
`DndContext` forwards its `id` prop into it. So the fix is to hand it an id that
does not come from the counter, and `useId` is React's answer to exactly this —
it derives one from the component's position in the tree, which is the same on
both sides however many requests the process has served.

```tsx
const dndContextId = useId()
// ...
<DndContext id={dndContextId} … >
```

## Why not expose it as a prop

We could also expose the `DndContext` id on `IlamyCalendarProps`, and
that would work — but it would only fix consumers who already know why they need
it, which is nobody until they have debugged this. Everyone who server-renders
hits it, and the fix costs one hook, so doing it internally means people get it
by upgrading. Happy to add the prop as well if you want an escape hatch, but I
could not find a case that needs one.

## Test

Renders the calendar three times in one process — which is what three successive
requests do — and asserts the described-by ids do not move. It needs an event on
screen: the attribute lives on draggables, so an empty calendar would pass
whether or not this were fixed. Verified by removing the `id` and watching it go
red.

## AI disclaimer

I've used Claude Opus 5 to help with this PR.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

### DIFF
--- a/packages/calendar/src/components/drag-and-drop/calendar-dnd-context.ssr.test.tsx
+++ b/packages/calendar/src/components/drag-and-drop/calendar-dnd-context.ssr.test.tsx
@@ -1,0 +1,52 @@
+import { describe, expect, test } from 'bun:test'
+import type { CalendarEvent } from '@ilamy/types'
+import dayjs from '@ilamy/utils/dayjs'
+import { renderToString } from 'react-dom/server'
+import { IlamyCalendar } from '@/features/calendar/components/ilamy-calendar'
+
+/*
+ * dnd-kit numbers the ids behind `aria-describedby` from a module-level
+ * counter. A browser gets a fresh module per page load; a server does not, so
+ * the counter carries from one request to the next and the markup drifts away
+ * from what the client will produce. Rendering twice in one process is exactly
+ * what two successive requests do, so it is what this asserts.
+ */
+describe('CalendarDndContext server rendering', () => {
+	const events: CalendarEvent[] = [
+		{
+			id: 'a',
+			title: 'A',
+			start: dayjs('2025-03-31T10:00:00.000Z'),
+			end: dayjs('2025-03-31T11:00:00.000Z'),
+		},
+	]
+
+	const renderOnce = () =>
+		renderToString(
+			<IlamyCalendar
+				events={events}
+				initialDate={dayjs('2025-03-31')}
+				initialView="week"
+			/>
+		)
+
+	// Draggables are what carry the attribute, so the calendar needs an event
+	// on screen for this to be testing anything at all.
+	const describedByIds = (html: string) => [
+		...new Set(
+			[...html.matchAll(/aria-describedby="([^"]*)"/g)].map((match) =>
+				match.at(1)
+			)
+		),
+	]
+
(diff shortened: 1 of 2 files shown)


kcsujeet (approved):
Thank you for the PR. LGTM

PR #264, opened 2026-09-02
feat: tell renderEvent which slice of an event it is drawing

### LINKED ISSUE #123: The "now" line renders in day-resolution vertical views, likely not needed

### LINKED ISSUE #263: renderEvent cannot tell a real event end from a week-boundary break
## What happens

An event too long for the row it starts in is cut at the boundary and drawn as
one bar per row. The built-in event content handles that: it squares the cut
edge, draws a continuation marker on it, and pads the title away from it.

A custom `renderEvent` is handed only the event, so it can do none of this. It
rounds both ends of both halves and marks neither, and one booking reads as two
bookings that happen to sit either side of a Sunday.

## Why there is no workaround

`isTruncatedStart` / `isTruncatedEnd` are computed in
`packages/calendar/src/lib/layout/horizontal.ts` and passed as props to
`DraggableEvent`. They never reach the context, the public API, or
`renderEvent`. Nothing a consumer can call reproduces them: the same
`CalendarEvent` object is handed to both halves, so there is nothing in the
argument to tell them apart.

The result is an asymmetry rather than a preference — the built-in renderer can
draw something a custom one cannot, and the moment you supply `renderEvent` you
lose it silently.

## Reproduce

```tsx
<IlamyCalendar
  initialView="month"
  initialDate={dayjs('2025-03-01')}
  events={[{ id: '1', title: 'Long booking',
             start: dayjs('2025-03-01T09:00Z'), end: dayjs('2025-03-04T17:00Z') }]}
  renderEvent={(event) => <div className="rounded-md bg-blue-500">{event.title}</div>}
/>
```

Saturday-to-Tuesday over a Monday-start week: two bars, both fully rounded,
neither marked as continuing. Removing `renderEvent` shows what the default
does instead.

## Related

The same call gives a custom renderer no way to place content against a
particular *day* of a multi-day bar, because it does not know how many days the
bar covers. `packages/calendar/src/lib/layout/horizontal.ts` computes exactly
that (`spanUnits = endCol - startCol + 1`) to derive `width`, then discards it.

## Version

`main` at b5b4514. Present since custom event rendering was introduced.

habovh:
## The gap

`DefaultEventContent` reads `isTruncatedStart` / `isTruncatedEnd` for four
separate things:

```tsx
getBorderRadiusClass(isTruncatedStart, isTruncatedEnd)          // square edge at a break
{isTruncatedStart && <div className="…bg-foreground/25" />}     // left continuation marker
{isTruncatedEnd   && <div className="…bg-foreground/25" />}     // right continuation marker
cn(isTruncatedStart && 'pl-1', isTruncatedEnd && 'pr-1')        // padding for those markers
```

And the call two lines below it:

```tsx
{renderEvent ? renderEvent(event) : <DefaultEventContent />}
```

So supplying `renderEvent` silently drops all four. The flags never leave
`lib/layout/horizontal.ts` — not into the context, not into the public API — and
both halves of a cut event are handed the same `CalendarEvent`, so there is
nothing in the argument to tell them apart. The built-in renderer can draw
something a custom one cannot.

## The change

```tsx
renderEvent(event, { isTruncatedStart, isTruncatedEnd, spanUnits })
```

Additive. A one-argument renderer is still assignable and still works — there is
a test that pins exactly that, because it is the property that makes this safe
to take.

An object rather than more positional parameters, so it can grow without another
signature change.

## About `spanUnits`

It grows once already, and I want to be straight about which half of this is
yours and which is mine.

The truncation flags are a gap in your API: your own renderer needs them, and
consumers cannot get them. That case stands on its own.

`spanUnits` is a need I hit downstream. A multi-day bar is one element spanning
several day cells, and one cell is `100 / spanUnits` percent of it — that is the
only way to place content against a particular day. Without it a start time can
only sit at the far end of the span, where it reads as the time the booking
*finishes*. With it, the start time lands over the day the booking starts and
the finish over the day it ends.

The layout already computes it, on the line that derives `width`, and then
throws it away:

```ts
const spanUnits = endCol - startCol + 1
// …
width: (spanUnits / bounds.unitCount) * 100,
```

So the cost is one field on `HorizontalPositionedEvent` and one prop. But it is
a smaller and more speculative case than the flags, and I would rather you knew
that than found it out in review. **If you would rather take the truncation
flags alone, say so and I will drop `spanUnits` — the shape is designed so it
can be added later without touching the signature again.**

## Tests

Three, on a Saturday-to-Tuesday booking over a Monday-start week, so the month
grid genuinely cuts it:

- each bar is told whether it is cut, and on which side — `start-real/end-cut`
  then `start-cut/end-real`
- each bar reports the units *it* covers (2 and 2), not the booking's four
- a one-argument renderer still works

Verified by handing `renderEvent` fixed values instead of the real segment and
watching the first two go red.

## Verification

`bun run check`, `bun run --filter '@ilamy/calendar' type-check` and
`bun run test` (1206 passing across all packages) are clean. The 17
`@ilamy/website` type errors are unchanged and present on an unmodified `main`.

## Note

Third PR from me, after the cross-month cells one and the SSR id one. They do
not conflict and can land in any order.

## Linked issue

Closes #263

## IA disclaimer

I used Claude Opus 5 to help work on this PR.

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

### DIFF
--- a/packages/calendar/src/components/draggable-event/draggable-event.tsx
+++ b/packages/calendar/src/components/draggable-event/draggable-event.tsx
@@ -30,6 +30,7 @@ function DraggableEventUnmemoized({
 	disableDrag = false,
 	isTruncatedStart = false,
 	isTruncatedEnd = false,
+	spanUnits = 1,
 	sourceResourceId,
 }: {
 	elementId: string
@@ -46,6 +47,11 @@ function DraggableEventUnmemoized({
 	/** Set by the horizontal events layer when the bar continues past the visible range. */
 	isTruncatedStart?: boolean
 	isTruncatedEnd?: boolean
+	/**
+	 * Grid units this bar covers, from the horizontal layout. One for a
+	 * vertical (time-column) event, which occupies a single day.
+	 */
+	spanUnits?: number
 }) {
 	const { onEventClick, renderEvent, disableEventClick, disableDragAndDrop } =
 		useSmartCalendarContext()
@@ -123,8 +129,18 @@ function DraggableEventUnmemoized({
 			{...attributes}
 			{...listeners}
 		>
-			{/* Use custom renderEvent from context if available, otherwise use default */}
-			{renderEvent ? renderEvent(event) : <DefaultEventContent />}
+			{/*
+				Use custom renderEvent from context if available, otherwise use
+				default. The segment travels with the event because the default
+				content above reads all of it — the border radius, both
+				continuation markers and their padding — and a custom renderer
+				that cannot see it has no way to draw the same thing.
+			*/}
+			{renderEvent ? (
+				renderEvent(event, { isTruncatedStart, isTruncatedEnd, spanUnits })
+			) : (
+				<DefaultEventContent />
+			)}
 		</AnimatedSection>
 	)
 }
@@ -139,7 +155,8 @@ export const DraggableEvent = memo(
 			prevProps.className === nextProps.className &&
 			prevProps.event === nextProps.event &&
 			prevProps.isTruncatedStart === nextProps.isTruncatedStart &&
-			prevProps.isTruncatedEnd === nextProps.isTruncatedEnd
+			prevProps.isTruncatedEnd === nextProps.isTruncatedEnd &&
+			prevProps.spanUnits === nextProps.spanUnits
 		)
 	}
 )
--- a/packages/calendar/src/components/draggable-event/render-event-segment.test.tsx
+++ b/packages/calendar/src/components/draggable-event/render-event-segment.test.tsx
@@ -1,0 +1,83 @@
+import { describe, expect, mock, test } from 'bun:test'
+import type { CalendarEvent } from '@ilamy/types'
+import dayjs from '@ilamy/utils/dayjs'
+import { cleanup, render, screen } from '@testing-library/react'
+import { CalendarProvider } from '@/features/calendar/contexts/calendar-context/provider'
+import type { EventSegment } from '@/features/calendar/types'
+import { MonthView } from '@/testing/view-harnesses'
+
+/*
+ * The built-in event content reads the whole segment — border radius, both
+ * continuation markers, and the padding that keeps text off them. A custom
+ * renderer that cannot see it has no way to draw the same thing, so every one
+ * of those affordances is lost the moment a consumer supplies `renderEvent`.
+ */
+describe('renderEvent segment', () => {
+	// A booking from Saturday to the following Tuesday, so the month grid cuts
+	// it at the week boundary and draws it as two bars.
+	const acrossTheWeekend: CalendarEvent[] = [
+		{
+			id: 'long',
+			title: 'Long booking',
+			start: dayjs('2025-03-01T09:00:00.000Z'),
+			end: dayjs('2025-03-04T17:00:00.000Z'),
+		},
+	]
+
+	const renderWithRenderer = (
+		renderEvent: (
+			event: CalendarEvent,
+			segment: EventSegment
+		) => React.ReactNode
+	) => {
+		cleanup()
+		return render(
+			<CalendarProvider
+				dayMaxEvents={5}
+				events={acrossTheWeekend}
+				firstDayOfWeek={1}
+				initialDate={dayjs('2025-03-01T00:00:00.000Z')}
+				renderEvent={renderEvent}
+			>
+				<MonthView />
+			</CalendarProvider>
+		)
+	}
+
+	const segmentsSeen = () =>
+		screen.getAllByTestId('segment').map((node) => node.textContent)
+
+	test('tells each bar whether it is cut, and on which side', () => {
+		renderWithRenderer((_event, segment) => (
+			<span data-testid="segment">
+				{segment.isTruncatedStart ? 'start-cut' : 'start-real'}/
+				{segment.isTruncatedEnd ? 'end-cut' : 'end-real'}
+			</span>
+		))
+
+		// The first bar really starts but is cut by the boundary; the second
+		// resumes and really ends.
+		expect(segmentsSeen()).toEqual(['start-real/end-cut', 'start-cut/end-real'])
+	})
+
+	test('reports the units each bar covers, not the whole booking', () => {
+		renderWithRenderer((_event, segment) => (
+			<span data-testid="segment">{segment.spanUnits}</span>
+		))
+
+		// Saturday and Sunday, then Monday and Tuesday — never four.
+		expect(segmentsSeen()).toEqual(['2', '2'])
+	})
+
+	test('leaves a one-argument renderer working', () => {
+		// The signature is additive: existing consumers pass a function that
+		// ignores the second argument, and nothing about them changes.
+		const oneArgument = mock((event: CalendarEvent) => (
+			<span data-testid="segment">{event.title}</span>
+		))
+		renderWithRenderer(oneArgument)
+
+		expect(segmentsSeen()).toEqual(['Long booking', 'Long booking'])
+		expect(oneArgument).toHaveBeenCalled()
+	})
+})
--- a/packages/calendar/src/features/calendar/types/index.ts
+++ b/packages/calendar/src/features/calendar/types/index.ts
@@ -22,6 +22,33 @@ export type SlotDuration = 15 | 30 | 60
  * Custom class names for calendar styling.
  * Allows users to override default styles for various calendar elements.
  */
+/**
+ * Which slice of an event a `renderEvent` call is drawing.
+ *
+ * An event too long for the row it starts in is cut at the boundary and drawn
+ * as one bar per row, each through its own `renderEvent` call. Without knowing
+ * which slice it has, a custom renderer cannot tell a real end from a break —
+ * so it rounds both, marks neither, and two halves of one booking read as two
+ * bookings that happen to sit either side of a Sunday. The built-in renderer
+ * has always had this; a custom one had no way to ask.
+ */
+export interface EventSegment {
+	/** The event starts before this bar; it is a continuation. */
+	isTruncatedStart: boolean
+	/** The event runs past this bar and resumes in the next row. */
+	isTruncatedEnd: boolean
+	/**
+	 * Grid units this bar covers — days in a day grid, hours in an hour one.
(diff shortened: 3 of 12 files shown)


packages/calendar/src/components/draggable-event/draggable-event.tsx
```diff
@@ -30,6 +30,7 @@ function DraggableEventUnmemoized({
 	disableDrag = false,
 	isTruncatedStart = false,
 	isTruncatedEnd = false,
+	spanUnits = 1,
```
kcsujeet: **issue:**
Could we pass real values from the vertical layer, or leave the fields out where we cannot fill them? Only `horizontal-grid-events-layer.tsx:90` supplies them today, so `resourceMonthVerticalColumns` (a day-resolution axis drawn through `VerticalGridEventsLayer`) hands a three-day bar `spanUnits: 1` and `isTruncatedStart: false`, where the horizontal layout gives 3 and true for the same event. FullCalendar sets its equivalents to `false` on cut-off bars rather than reporting them whole.

packages/calendar/src/features/calendar/types/index.ts
```diff
@@ -22,6 +22,33 @@ export type SlotDuration = 15 | 30 | 60
...
+ * An event too long for the row it starts in is cut at the boundary and drawn
+ * as one bar per row, each through its own `renderEvent` call. Without knowing
+ * which slice it has, a custom renderer cannot tell a real end from a break —
+ * so it rounds both, marks neither, and two halves of one booking read as two
+ * bookings that happen to sit either side of a Sunday. The built-in renderer
+ * has always had this; a custom one had no way to ask.
+ */
+export interface EventSegment {
+	/** The event starts before this bar; it is a continuation. */
+	isTruncatedStart: boolean
```
kcsujeet: **suggestion:**
Could we call these `isStart` and `isEnd` on the public type, inverting the values at the one call site? FullCalendar hands its renderers the same two facts under those names with the opposite polarity, so a consumer porting an `eventContent` writes `isStart`, gets no type error, and rounds the wrong corners. The internal props and the layout type can keep the names they have.

FullCalendar's definition, from `packages/preact/src/component-util/event-rendering.ts:97-98` (the props are listed on https://fullcalendar.io/docs/event-render-hooks):

```ts
isStart: normalRange.start && normalRange.start.valueOf() === slicedRange.start.valueOf(),
isEnd:   normalRange.end   && normalRange.end.valueOf()   === slicedRange.end.valueOf(),
```

So `isStart` is true when the bar holds the event's real start, which is exactly `!isTruncatedStart`. The same renderer written each way:

```tsx
// FullCalendar
eventContent: (info) => <div className={info.isStart ? 'rounded-l' : 'square-l'} />

// here, today
renderEvent: (event, segment) => <div className={segment.isTruncatedStart ? 'square-l' : 'rounded-l'} />

// here, with the rename
renderEvent: (event, segment) => <div className={segment.isStart ? 'rounded-l' : 'square-l'} />
```

The type and the call site:

```ts
// features/calendar/types/index.ts
export interface EventSegment {
	/** This bar holds the event's real start. */
	isStart: boolean
	/** This bar holds the event's real end. */
	isEnd: boolean
}
```

```tsx
// draggable-event.tsx, the single boundary where the flip happens
const segment: EventSegment = {
	isStart: !isTruncatedStart,
	isEnd: !isTruncatedEnd,
}
```

packages/calendar/src/features/calendar/types/index.ts
```diff
@@ -22,6 +22,33 @@ export type SlotDuration = 15 | 30 | 60
...
+	/**
+	 * Grid units this bar covers — days in a day grid, hours in an hour one.
+	 * One for a vertical (time-column) event, which occupies a single day.
+	 *
+	 * What it buys you is the width of a single unit, `100 / spanUnits` percent
+	 * of the bar, which is the only way to place content against a particular
+	 * day of a multi-day bar: a start time over the day it starts on rather
+	 * than floating at the far end of the span, where it reads as a finish.
+	 */
+	spanUnits: number
```
kcsujeet: **suggestion:**
Lean toward landing the two flags now and leaving `spanUnits` until something needs it. Nothing reads it yet, and as a required field of an exported type it is additive to add later but breaking to remove. If you do want it, a start and end date for the slice would also say *which* days the bar covers, which a bare count cannot:

```ts
export interface EventSegment {
	isStart: boolean
	isEnd: boolean
	/** The slice this bar draws, clipped to the row it sits in. */
	start: Dayjs
	end: Dayjs
}
```

packages/calendar/src/components/draggable-event/draggable-event.tsx
```diff
@@ -123,8 +129,18 @@ function DraggableEventUnmemoized({
...
-			{/* Use custom renderEvent from context if available, otherwise use default */}
-			{renderEvent ? renderEvent(event) : <DefaultEventContent />}
+			{/*
+				Use custom renderEvent from context if available, otherwise use
+				default. The segment travels with the event because the default
+				content above reads all of it — the border radius, both
+				continuation markers and their padding — and a custom renderer
+				that cannot see it has no way to draw the same thing.
+			*/}
+			{renderEvent ? (
```
kcsujeet: **suggestion:**
Could we lift these into named consts above the return? That keeps the ternary on one line and gives the explanation a home other than a comment sitting on a branch.

```tsx
const segment: EventSegment = { isTruncatedStart, isTruncatedEnd, spanUnits }
const content = renderEvent ? renderEvent(event, segment) : <DefaultEventContent />
```

packages/calendar/src/features/calendar/types/index.ts
```diff
@@ -22,6 +22,33 @@ export type SlotDuration = 15 | 30 | 60
  * Custom class names for calendar styling.
  * Allows users to override default styles for various calendar elements.
  */
+/**
```
kcsujeet, line 21: **chore:**
Could we move the `EventSegment` block above line 21? It landed between `CalendarClassesOverride`'s comment and its declaration, so that interface now ships with no description and `EventSegment` carries two stacked comment blocks.

packages/calendar/src/components/draggable-event/render-event-segment.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+	test('reports the units each bar covers, not the whole booking', () => {
+		renderWithRenderer((_event, segment) => (
+			<span data-testid="segment">{segment.spanUnits}</span>
+		))
+
+		// Saturday and Sunday, then Monday and Tuesday — never four.
+		expect(segmentsSeen()).toEqual(['2', '2'])
+	})
+
+	test('leaves a one-argument renderer working', () => {
```
kcsujeet, line 66: **suggestion:**
Worth making this a type-level check instead. Whether a one-argument renderer still fits is something `tsc` decides, so reverting the call site to `renderEvent(event)` leaves this test green while the other two fail. `packages/utils/src/dayjs.test.ts:38` uses `@ts-expect-error` for the same kind of guarantee:

```ts
// The old one-argument renderer must keep fitting.
const oneArgument: NonNullable<IlamyCalendarProps['renderEvent']> = (event) =>
	event.title

// @ts-expect-error a third argument is not part of the contract
const threeArguments: NonNullable<IlamyCalendarProps['renderEvent']> = (
	event,
	segment,
	extra
) => null
```

packages/calendar/src/components/draggable-event/render-event-segment.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+		// The signature is additive: existing consumers pass a function that
+		// ignores the second argument, and nothing about them changes.
+		const oneArgument = mock((event: CalendarEvent) => (
+			<span data-testid="segment">{event.title}</span>
+		))
+		renderWithRenderer(oneArgument)
+
+		expect(segmentsSeen()).toEqual(['Long booking', 'Long booking'])
+		expect(oneArgument).toHaveBeenCalled()
```
kcsujeet: **nitpick:**
Worth `toHaveBeenCalledTimes(2)` here, since the event draws as two bars and the count is known.

packages/calendar/src/index.ts
```diff
@@ -44,6 +44,7 @@ export {
 export type {
 	CalendarClassesOverride,
 	CellInfo,
+	EventSegment,
```
kcsujeet, line 47: **chore:**
Could we update the docs alongside this? `apps/website/src/content/docs/components/calendar.mdx:82` and `docs/types-and-interfaces.md:113` still show `renderEvent` taking only an event, and `EventSegment` is missing from the type catalogue.

packages/calendar/src/components/draggable-event/draggable-event.tsx
```diff
@@ -139,7 +155,8 @@ export const DraggableEvent = memo(
 			prevProps.className === nextProps.className &&
 			prevProps.event === nextProps.event &&
 			prevProps.isTruncatedStart === nextProps.isTruncatedStart &&
-			prevProps.isTruncatedEnd === nextProps.isTruncatedEnd
+			prevProps.isTruncatedEnd === nextProps.isTruncatedEnd &&
```
kcsujeet: **nitpick:**
Worth switching this to a key list with `every`, which would also surface that `style` and `sourceResourceId` are never compared. Each new prop currently costs another clause in the chain.

kcsujeet (approved):
Approving with the review findings applied on the branch rather than sent back for another round.

Changed on takeover: the segment fields are now \`isStart\` / \`isEnd\`, matching FullCalendar's event render hook, where they mean the segment's range matches the event's own range. The previous names described the cut, so every caller inverted them at the point of use.

Day and week columns now report truncation too. \`computeTopHeight\` already clamped an event running past the visible hours; it just never said so, and an overnight bar in a 09:00-17:00 grid claimed both edges were real. This also changes the default bar, which reads the same flags.

Also dropped \`spanUnits\` from the contract, replaced a test that could not fail with a type-level one, and updated the prop tables with a worked example.

Thanks for this. You are kept as author and co-author on the commit.

Issue #247, opened 2026-08-12
[Bug] ISO strings with a `Z` offset are parsed as local wall-clock time


Paul6552:
### Description

  `IlamyCalendar` accepts `Dayjs | Date | string` for an event's `start` and
  `end` properties. However, ISO strings containing a UTC offset are parsed
  incorrectly when the calendar runs in a non-UTC timezone.

  For example, in Europe/Vienna:

```
  const events = [{
    id: 1,
    title: 'All-day event',
    start: '2026-08-17T22:00:00.000Z',
    end: '2026-08-18T22:00:00.000Z',
    allDay: true,
  }];
```

  2026-08-17T22:00:00.000Z represents 2026-08-18T00:00:00+02:00.
  However, ilamy interprets it as 2026-08-17T22:00:00+02:00, causing the
  event to appear on August 17 as well.

  ### Minimal reproduction

```
  import dayjs from 'dayjs';
  import { dayjs as ilamyDayjs } from '@ilamy/calendar';

  const value = '2026-08-17T22:00:00.000Z';

  console.log(dayjs(value).format());
  // 2026-08-18T00:00:00+02:00

  console.log(ilamyDayjs(value).format());
  // 2026-08-17T22:00:00+02:00
```

  Environment timezone: Europe/Vienna

  Passing the same instant as a Date works correctly:

```
  ilamyDayjs(new Date(value)).format();
  // 2026-08-18T00:00:00+02:00
```

  ### Expected behavior

  ISO strings should preserve their represented instant and offset, just as
  standard Dayjs parsing and JavaScript Date do.

  ### Actual behavior

  The UTC timestamp is treated as if its clock components were already in the
  local timezone.

  ### Environment

  - @ilamy/calendar: 2.0.3
  - dayjs: 1.11.21
  - Browser/system timezone: Europe/Vienna

kcsujeet:
Fixed in v3.0.0 — https://github.com/kcsujeet/ilamy-calendar/releases/tag/v3.0.0

Thanks for trying ilamy calendar! Feel free to open more issues any time.

PR #266, opened 2026-09-02
fix: open on the given date, read on the calendar's clock

### LINKED ISSUE #123: The "now" line renders in day-resolution vertical views, likely not needed

### LINKED ISSUE #265: Calendar opens on the wrong day when timezone differs from the browser's
## What happens

With a `timezone` set, the calendar can open on the day *before* (or after) the
one it was given. A day view asked for 10 March shows "Sunday, March 9".

It only happens for readers whose machine is in a different zone from the
`timezone` prop — which is the reader the prop exists for. Anyone testing from a
machine that agrees with the workspace sees nothing wrong, which is why this
survived #247.

## Reproduce

```tsx
// machine in Europe/Paris (UTC+2)
<IlamyCalendar initialDate="2025-03-10" initialView="day" timezone="UTC" />
```

The header reads **Sunday, March 9, 2025**. `currentDate` is
`2025-03-09 23:00 +00:00`.

With the machine on UTC the same code gives Monday, March 10 — so the output
depends on where the browser is, which is the thing `timezone` is supposed to
settle.

Confirmed on `main` at b5b4514 with a `TZ=Europe/Paris` test run.

## Cause

`initialDate` is parsed before the calendar's zone is known — by the consumer's
own module, or by `safeDate` in `IlamyCalendar` — so it is a moment in the
*machine's* zone. `useCalendarNavigation` then seeds state with
`dayjs(initialDate)`, which cannot re-anchor it; its own comment notes this is
"the last place that can fix that", but nothing applies the zone there.

The timezone effect in `useCalendarEngine` later runs
`setCurrentDate((prev) => prev.tz(timezone))`, which preserves the *instant*.
That is right for events — an event is a moment — but the current date is a
position in the calendar, and preserving its instant moves it across midnight.

## Note on the two cases

They want different answers, which is what makes this more than a one-liner:

- A date **given** is a calendar date. `2025-03-10` should open on 10 March
  wherever it is read.
- **No** date is "now" — a moment. It should resolve to the calendar's today,
  not the machine's; near midnight those differ.

Treating both as wall clocks fixes the first and breaks the second.

## Version

`main` at b5b4514.

habovh:
## The bug

With `timezone` set, the calendar opens on the wrong day for any reader whose
machine is in a different zone — the reader the prop exists for.

```tsx
// machine in Europe/Paris
<IlamyCalendar initialDate="2025-03-10" initialView="day" timezone="UTC" />
// header: "Sunday, March 9, 2025";  currentDate: 2025-03-09 23:00 +00:00
```

`initialDate` is parsed before the zone is known, so it is a moment in the
machine's zone by the time the engine sees it. The timezone effect then converts
it with `prev.tz(timezone)`, which preserves the instant — correct for an event,
which *is* a moment, but the current date is a position in the calendar, and
preserving its instant walks it across midnight.

`useCalendarNavigation`'s own comment already says it is "the last place that
can fix that, since nothing downstream re-parses it" — this puts the zone there.

## The change

One helper at the point the date enters the engine, and the two cases answered
separately:

```ts
if (!timezone) return initialDate ?? dayjs()
// no date is "now": a moment, kept as an instant and read on the calendar's clock
if (!initialDate) return dayjs().tz(timezone)
// a date given is a calendar date: its wall clock kept, read in the calendar's zone
return dayjs.tz(initialDate.format('YYYY-MM-DDTHH:mm:ss'), timezone)
```

The single rule — treat both as wall clocks — is tempting and wrong: a
workspace in Kiritimati would open on the reader's yesterday. The second test
below fails under it, which is why it is there.

Anchoring during render is safe. `dayjs.tz` with an explicit zone is a pure
call, unlike the `setDefault` whose module-global write is exactly why the
effect underneath it has to wait for mount — so this does not disturb the SSR
reasoning in that comment.

The effect is untouched, and on mount its conversion is now a no-op, since the
date is already in the zone.

## Tests

Two, in the existing `IlamyCalendar - timezone prop` block from #247, using
Kiritimati (+14) and Midway (−11) so neither passes by accident on any machine:

- a given date opens on that date in both zones
- with no date, the calendar opens on *its* today, not the machine's

Verified by reverting: the first goes red without the fix, and the second goes
red under the naive single-rule version.

## Verification

`bun run check`, `bun run --filter '@ilamy/calendar' type-check` and
`bun run test` (1205 passing) are clean. The 17 `@ilamy/website` type errors are
unchanged and present on an unmodified `main`.

## Note

Fourth PR from me, after the cross-month cells, the SSR id, and the
`renderEvent` segment one. None of them conflict and they can land in any order.

## Linked issue

Closes #265 

## IA disclaimer

I used Claude Opus 5 to help on this PR.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Maintenance

## Checklist

- [x] Changes are tested

### DIFF
--- a/packages/calendar/src/features/calendar/components/ilamy-calendar.test.tsx
+++ b/packages/calendar/src/features/calendar/components/ilamy-calendar.test.tsx
@@ -11,12 +11,21 @@ import {
 import { recurrencePlugin } from '@ilamy/calendar-recurrence'
 import type { CalendarEvent, IlamyPlugin } from '@ilamy/types'
 import dayjs from '@ilamy/utils/dayjs'
-import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import {
+	cleanup,
+	fireEvent,
+	render,
+	screen,
+	waitFor,
+} from '@testing-library/react'
 import { createContext, useContext } from 'react'
 import { RRule } from 'rrule'
 import type { EventFormProps } from '@/features/calendar/components/event-form/event-form'
 import { useIlamyCalendarContext } from '@/features/calendar/hooks/use-smart-calendar-context'
-import type { IlamyCalendarPropEvent } from '@/features/calendar/types'
+import type {
+	IlamyCalendarPropEvent,
+	IlamyCalendarProps,
+} from '@/features/calendar/types'
 import { DISABLED_CELL_CLASSNAME } from '@/lib/constants'
 import { IlamyCalendar } from './ilamy-calendar'
 
@@ -1239,6 +1248,53 @@ describe('IlamyCalendar - timezone prop', () => {
 		const parsedBeforeMount = dayjs('2025-01-15T12:00:00.000Z')
 
 		expect(renderWithTokyo(parsedBeforeMount)).toBe('21:00 +09:00')
+	})
+
+	/*
+	 * The same question asked of the date the calendar OPENS on. `initialDate`
+	 * is parsed before the zone is known, so the calendar was showing whichever
+	 * day that date fell on in the machine's zone — off by one for any reader
+	 * far enough from the workspace, which is the case the prop exists for.
+	 *
+	 * Two zones either side of every machine, so neither passes by accident.
+	 */
+	const DayProbe = () => (
+		<span data-testid="probe-day">
+			{useIlamyCalendarContext().currentDate.format('YYYY-MM-DD')}
+		</span>
+	)
+
+	const dayShownFor = (
+		timezone: string,
+		initialDate?: IlamyCalendarProps['initialDate']
+	) => {
+		cleanup()
+		render(
+			<IlamyCalendar
+				headerComponent={<DayProbe />}
+				initialDate={initialDate}
+				initialView="day"
+				timezone={timezone}
+			/>
+		)
+		return screen.getByTestId('probe-day').textContent
+	}
+
+	it('opens on the date it was given, wherever the reader is', () => {
+		// A date, not a moment: the consumer named 10 March, so 10 March is what
+		// opens — in Kiritimati (+14) and in Midway (-11) alike.
+		expect(dayShownFor('Pacific/Kiritimati', '2025-03-10')).toBe('2025-03-10')
+		expect(dayShownFor('Pacific/Midway', '2025-03-10')).toBe('2025-03-10')
+	})
+
+	it("opens on the calendar's today when given no date", () => {
+		// No date is a different question: "now", answered on the calendar's
+		// clock rather than the machine's. Near midnight those are different days.
+		for (const timezone of ['Pacific/Kiritimati', 'Pacific/Midway']) {
+			expect(dayShownFor(timezone)).toBe(
+				dayjs().tz(timezone).format('YYYY-MM-DD')
+			)
+		}
 	})
 })
 
--- a/packages/calendar/src/features/calendar/hooks/use-calendar-engine.ts
+++ b/packages/calendar/src/features/calendar/hooks/use-calendar-engine.ts
@@ -101,6 +101,40 @@ interface CalendarEngineHandlers {
 	handleDateClick: (info: CellInfo) => void
 }
 
+/**
+ * The date the calendar opens on, read on the calendar's clock rather than the
+ * machine's.
+ *
+ * `initialDate` is parsed by the consumer's own module, or by `safeDate` in
+ * `IlamyCalendar`, and both run before the zone is known — so the calendar was
+ * opening on whichever day that date fell on where the browser happens to be.
+ * For a reader far enough from the workspace that is the wrong day, which is
+ * exactly the reader `timezone` exists for.
+ *
+ * The two cases are different questions and get different answers:
+ *
+ * - A date GIVEN is a calendar date. The consumer wrote 10 March; 10 March is
+ *   what opens. Its wall clock is kept and read in the calendar's zone.
+ * - NO date is "now". That is a moment, so the instant is kept and read on the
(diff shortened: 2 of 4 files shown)


packages/calendar/src/features/calendar/hooks/use-calendar-engine.ts
```diff
@@ -101,14 +101,48 @@ interface CalendarEngineHandlers {
...
+	initialDate: Dayjs | undefined,
+	timezone: string | undefined
+): Dayjs => {
+	if (!timezone) {
+		return initialDate ?? dayjs()
+	}
+	if (!initialDate) {
+		return dayjs().tz(timezone)
+	}
+	return dayjs.tz(initialDate.format('YYYY-MM-DDTHH:mm:ss'), timezone)
```
kcsujeet: **issue:**
Could we branch on whether the value carries its own instant, rather than on whether a date was given? `safeDate` has already parsed the prop into a Dayjs in the machine's zone by the time this runs, so reading its clock face and re-anchoring moves the instant and makes the opening day depend on the browser. With `timezone="Pacific/Kiritimati"` and `initialDate='2025-03-10T22:00:00.000Z'` (the true day there is 2025-03-11) I get 2025-03-10 on a UTC machine, 2025-03-11 on Tokyo and 2025-03-10 on New York, and a `Date` or a consumer `dayjs()` behaves the same.

`timezoneAwareDayjs` (`packages/utils/src/dayjs.ts:172-190`) already makes this exact call, so the rule is settled and only the plumbing is open: it reads the zone from the module global that `setDefault` has not written during the first render, and `safeDate` has erased the string by the time the engine could ask.

packages/calendar/src/features/calendar/components/ilamy-calendar.test.tsx
```diff
@@ -1240,6 +1249,53 @@ describe('IlamyCalendar - timezone prop', () => {
...
+		)
+		return screen.getByTestId('probe-day').textContent
+	}
+
+	it('opens on the date it was given, wherever the reader is', () => {
+		// A date, not a moment: the consumer named 10 March, so 10 March is what
+		// opens — in Kiritimati (+14) and in Midway (-11) alike.
+		expect(dayShownFor('Pacific/Kiritimati', '2025-03-10')).toBe('2025-03-10')
+		expect(dayShownFor('Pacific/Midway', '2025-03-10')).toBe('2025-03-10')
+	})
```
kcsujeet, line 1295: **suggestion:**
Worth adding a `Date` or `Dayjs` case here, plus one `toISOString()` assertion. The only `initialDate` these tests supply is `'2025-03-10'`, so the shapes that break above go untested, and asserting only `format('YYYY-MM-DD')` lets a shifted instant hide behind a matching format.

packages/calendar/src/features/calendar/hooks/use-calendar-engine.ts
```diff
@@ -101,14 +101,48 @@ interface CalendarEngineHandlers {
...
+}
```
kcsujeet, line 134: **chore:**
Could we update the places that still name the navigation slice as where `initialDate` gets anchored? `docs/timezones.md` (both the "Incoming props are re-anchored" section and the Key files list), `packages/utils/src/helpers.ts:7-9` and `use-calendar-navigation.ts:62-65` all describe the old order now that the engine anchors first.

packages/calendar/src/features/calendar/components/ilamy-calendar.test.tsx
```diff
@@ -1240,6 +1249,53 @@ describe('IlamyCalendar - timezone prop', () => {
...
+
+	it("opens on the calendar's today when given no date", () => {
+		// No date is a different question: "now", answered on the calendar's
+		// clock rather than the machine's. Near midnight those are different days.
+		for (const timezone of ['Pacific/Kiritimati', 'Pacific/Midway']) {
+			expect(dayShownFor(timezone)).toBe(
+				dayjs().tz(timezone).format('YYYY-MM-DD')
+			)
```
kcsujeet: **suggestion:**
Could we pin the clock here and assert an absolute day? Recomputing `dayjs().tz(timezone).format('YYYY-MM-DD')` at assert time races midnight, since the component reads `dayjs()` during render and the expectation reads it again afterwards. `year-view.test.tsx:649-676` uses `setSystemTime` for this.

packages/calendar/src/features/calendar/hooks/use-calendar-engine.ts
kcsujeet: **suggestion:**
If the helper keeps taking a `Dayjs`, `initialDate.tz(timezone, true)` would do this without the format round trip. That is dayjs's own keep-local-time path, `packages/utils/src/dayjs.ts:51` already uses it, and the round trip drops milliseconds on the way through.

packages/calendar/src/features/calendar/hooks/use-calendar-engine.ts
```diff
@@ -149,7 +183,7 @@ export const useCalendarEngine = (
 	const pluginRuntime = useMemo(() => createPluginRuntime(plugins), [plugins])
 
 	const navigation = useCalendarNavigation({
-		initialDate,
+		initialDate: anchorInitialDate(initialDate, timezone),
```
kcsujeet: **nitpick:**
Worth wrapping this in `useState(() => ...)`, since `useCalendarNavigation` only reads `initialDate` in its lazy initializer (`use-calendar-navigation.ts:66`) and every call after mount is discarded.

PR #273, opened 2026-09-12
fix(header): animate a header label only when it has something to show

kcsujeet:
Fixes the header animations replaying on every navigation, whether or not the label had anything new to show.

## The problem

`AnimatedSection` is a CSS `animate-in` that replays whenever its node is remounted. Every header cell was keyed by `day.toISOString()`, so navigating replaced the nodes and re-ran the fade over text that had not changed.

Measured through the header's own Next button, before the fix:

```
PROBE month weekday: "Sun" -> "Sun" | same DOM node: false
PROBE week:          "Sun2" -> "Sun9" | same DOM node: false
```

The month case is the obvious one: every month shows the same seven names, so the entire weekday row blinked for nothing. In the week header the number genuinely moves, but the weekday was animated along with it, so `Sun` was re-introduced next to a number that had actually changed.

## The fix

Key each header cell by what it renders, not by the date that produced it.

- **Month weekday row** is keyed by the name. It animates once on mount and then sits still.
- **Week and resource day headers** are keyed by position, and the day number animates on its own via a new `AnimatedDayLabel`. The weekday stays put; the number moves.
- **Hour row** is keyed by the hour, which is identical whatever day is shown.
- **Day view header** is unchanged: its whole label (`dddd, LL`) really does change on navigation, so animating it is correct.

After:

```
PROBE month weekday: "Sun" -> "Sun"  | remounted: false
PROBE week: "Sun2" -> "Sun9" | weekday remounted: false | number subtree remounted: true
```

## Verification

Two tests added to `ilamy-calendar.test.tsx`, asserting DOM node identity across a real Next click rather than class names, since identity is what decides whether the animation plays. Both are load-bearing: restoring either old key fails its test.

`build`, `type-check` (0 errors), full suite (926 tests, 0 fail) and the Fallow gate all pass.

## Notes

Header cells are keyed by position deliberately. They are slots, not identities — the second column of a week header is Monday's whatever week is displayed — and the part that varies inside is keyed separately. The usual objection to index keys does not apply to a fixed-length list that is never reordered.

Worth recording: rerendering with a different `initialDate` does not navigate, because `initialDate` is read once per mount (the `useState` from #266). My first measurement did that and showed nothing changing. Navigation has to go through the header button, which is also the only way to reproduce what you actually see.


kcsujeet:
Pushed `fd3c12f`, which corrects the rule I got wrong.

You were right on both counts. My first pass froze the weekday and animated the number, which is only correct for a week column. A month column is the mirror image: the first column is day 1 of every month, and the weekday under it is what moves. So freezing the weekday left both resource month headers with no animation at all.

Measured through the Next button, March to April, first day label:

```
resource month vertical:   "Sat1" -> "Tue1" | weekday animated: true | number animated: false
resource month horizontal: "Sat1" -> "Tue1" | weekday animated: true | number animated: false
week column:               "Sun2" -> "Sun9" | weekday animated: false | number animated: true
```

`AnimatedDayLabel` now keys each half to its own text, so the fade plays on whichever half changed and no header needs per-view knowledge. That is the rule that survives contact with all five headers.

Two labels that were never animated at all now follow it too: the week number in the corner cell, and the resource month vertical gutter, which was still a plain `DayLabel`.

Tests are up to five, covering both directions of the rule, and each fails when its own half of the fix is reverted. 929 tests pass; build, type-check and Fallow clean.

